PR #27, opened 2015-01-20
Allow for zero

mejackreed:
Previously, https://github.com/projectblacklight/blacklight_range_limit/compare/allow-for-zero?expand=1#diff-f17504e7a61c53d7423849464de8136aL13 was checking for the truthiness of the min and max values. This was causing an error when a value was zero. In JavaScript, zero, `0`, evaluates as falsey.

https://dorey.github.io/JavaScript-Equality-Table/

3 original tests fail when a zero value document is indexed (js errors). Added in an easy way to index testing documents for development and testing in 878a66f.

Now tests for a value being an int.

### DIFF
--- a/lib/generators/blacklight_range_limit/blacklight_range_limit_generator.rb
+++ b/lib/generators/blacklight_range_limit/blacklight_range_limit_generator.rb
@@ -7,5 +7,4 @@ class BlacklightRangeLimitGenerator < Rails::Generators::Base
   def copy_public_assets
     BlacklightRangeLimit::AssetsGenerator.start
   end
-
 end
--- a/Rakefile
+++ b/Rakefile
@@ -40,6 +40,7 @@ namespace :test do
     within_test_app do
       ENV['RAILS_ENV'] ||= 'test'
       system "rake blacklight:solr:seed"
+      system "rake blacklight_range_limit:seed"
     end
   end
 
--- a/app/assets/javascripts/blacklight_range_limit/range_limit_slider.js
+++ b/app/assets/javascripts/blacklight_range_limit/range_limit_slider.js
@@ -10,7 +10,7 @@ $(".range_limit .profile .range.slider_js").each(function() {
    var min = boundaries[0];
    var max = boundaries[1];
 
-   if (min && max) {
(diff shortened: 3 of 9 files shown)


lib/generators/blacklight_range_limit/blacklight_range_limit_generator.rb
```diff
@@ -8,4 +8,14 @@ def copy_public_assets
     BlacklightRangeLimit::AssetsGenerator.start
   end
 
+  def fixtures
+    FileUtils.mkdir_p 'spec/fixtures/solr_documents'
```
jcoyne: I don't think this belongs in this generator. Maybe you want to make a test application generator?

lib/generators/blacklight_range_limit/blacklight_range_limit_generator.rb
```diff
@@ -8,4 +8,14 @@ def copy_public_assets
...
+    directory 'solr_documents', 'spec/fixtures/solr_documents'
+  end
+
+  def inject_into_catalog_controller
+    inject_into_file 'app/controllers/catalog_controller.rb', after: "config.add_facet_field 'example_pivot_field', :label => 'Pivot Field', :pivot => ['format', 'language_facet']" do
```
jcoyne: This may not succeed if the catalog controller has been changed. I don't think it belongs in this generator anyway.

lib/generators/blacklight_range_limit/blacklight_range_limit_generator.rb
mejackreed: Yep, thanks for pointing that out.

lib/generators/blacklight_range_limit/blacklight_range_limit_generator.rb
mejackreed: Your right, adding back to the test app generator.

mejackreed:
@jcoyne Updated with your comments... Also modified brittle helper specs which were not passing due to varying versions of BL/Bootstrap

jcoyne:
:+1: Looks good to me. Thanks @mejackreed for fixing the test suite too.

PR #32, opened 2015-03-25
Move #add_range_limit_params into Blacklight::Solr::SearchBuilder

rwd:
This moves the `add_range_limit_params` processor out of the controller and into `Blacklight::Solr::SearchBuilder`, following the move of search builder logic into its `Blacklight::SearchBuilder`.

This would however make blacklight_range_limit depend on blacklight version > 5.10.0, so the gemspec should be updated if this is accepted.

### DIFF
--- a/lib/blacklight_range_limit.rb
+++ b/lib/blacklight_range_limit.rb
@@ -2,6 +2,7 @@
 
 module BlacklightRangeLimit
   autoload :ControllerOverride, 'blacklight_range_limit/controller_override'
+  autoload :SearchBuilderOverride, 'blacklight_range_limit/search_builder_override'
   autoload :ViewHelperOverride, 'blacklight_range_limit/view_helper_override'
   autoload :RouteSets, 'blacklight_range_limit/route_sets'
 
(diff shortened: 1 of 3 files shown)


lib/blacklight_range_limit.rb
```diff
@@ -22,10 +23,14 @@ def self.omit_inject=(value)
   def self.omit_inject ; @omit_inject ; end
   
   def self.inject!
-    unless omit_inject[:controller_mixin]
+      unless omit_inject[:controller_mixin]
         CatalogController.send(:include, BlacklightRangeLimit::ControllerOverride) unless Blacklight::Catalog.include?(BlacklightRangeLimit::ControllerOverride)
       end
 
+      unless omit_inject[:search_builder_mixin]
+        Blacklight::Solr::SearchBuilder.send(:include, BlacklightRangeLimit::SearchBuilderOverride) unless Blacklight::Solr::SearchBuilder.include?(BlacklightRangeLimit::SearchBuilderOverride)
```
jcoyne: I think we should move this into a generator that adds this to `app/models/search_builder.rb`

lib/blacklight_range_limit/search_builder_override.rb
```diff
@@ -0,0 +1,53 @@
+module BlacklightRangeLimit
+  module SearchBuilderOverride
+    include SegmentCalculation
+
+    included do
```
jcoyne: this module must `extend ActiveSupport::Concern` in order to use `included`

cbeer:
There's also a branch at https://github.com/projectblacklight/blacklight_range_limit/compare/blacklight-master that I think is getting at the same thing.

lib/blacklight_range_limit.rb
cbeer: And just require BL `>= 5.12`?

jrochkind:
Handled by #37, with more deprecations and API changes since this PR. Sorry, I hadn't noticed yours before starting on mine!

mejackreed:
:sunglasses: all good

PR #33, opened 2015-04-07
Update to require Blacklight 5.10

mejackreed:
This looked good to me.... @cbeer Anything else you think that needs to go in here?

PR #36, opened 2015-07-10
fix issue with numbers that had commas getting truncated

peetucket:
also , fix a coupe deprecation warnings

note...could not run the tests


lib/blacklight_range_limit/controller_override.rb
```diff
@@ -8,7 +8,7 @@ module ControllerOverride
     extend ActiveSupport::Concern
   
     included do
-      solr_search_params_logic << :add_range_limit_params
```
cbeer: I think we'll want to release with the fixed number parsing without requiring Blacklight 5.10+ (which is done in #33)

app/assets/javascripts/blacklight_range_limit/range_limit_shared.js
```diff
@@ -0,0 +1,7 @@
+
+// takes a string and parses into an integer, but throws away commas first, to avoid truncation when there is a comma
+// use in place of javascript's native parseNum
+function parseNum(str) {
+  str = String(str).replace(/,/g, "");
```
cbeer: Couple comments:

- I guess this is good enough, with the caveat that it won't work for some other Rails locales. I guess the right way to fix this is (in Blacklight 6?) including a `data-` attribute for just the number and use that.

Another thing we could do is just extract the digits, instead of trying to deal with separators.

- Do we really want to pollute the global namespace with this function? I guess the other `blacklight_range_limit` functions are all anonymous, but maybe we should introduce a class? @mejackreed?

VERSION
```diff
@@ -1 +1 @@
-5.0.4
```
cbeer: It would be better not to bump the version in the same commit as the fix -- it'll make it easier to backport later if the fix is better isolated.

app/assets/javascripts/blacklight_range_limit/range_limit_shared.js
mejackreed: Agreed about the global scope here. I think scoping under  `BlacklightRangeLimit` probably makes the most sense here. Here are two approaches:

### Using [Revealing Module Pattern](http://addyosmani.com/resources/essentialjsdesignpatterns/book/#revealingmodulepatternjavascript)

```javascript
var BlacklightRangeLimit = (function(global) {

  var parseNum = function(str) {
    str = String(str).replace(/,/g, "");
    return parseInt(str);
  };
  
  return {
   parseNum: parseNum
  };

})();
```

### Using [Airbnb Javascript Style Guide Module Pattern](https://github.com/airbnb/javascript/tree/master/es5#modules)
```javascript
!function(global) {
  'use strict';

  var previousBlacklightRangeLimit = global.BlacklightRangeLimit;

  function BlacklightRangeLimit(options) {
    this.options = options || {};
  }

  BlacklightRangeLimit.noConflict = function noConflict() {
    global.BlacklightRangeLimit = previousBlacklightRangeLimit;
    return BlacklightRangeLimit;
  };

  BlacklightRangeLimit.parseNum = function parseNum(str) {
    str = String(str).replace(/,/g, "");
    return parseInt(str);
  };

  global.BlacklightRangeLimit = BlacklightRangeLimit;
}(this);
```

The first approach is simpler, the second one checks if the module is already defined.

VERSION
peetucket: Will rollback version number to 5.0.4 in next commit.

app/assets/javascripts/blacklight_range_limit/range_limit_shared.js
peetucket: Went with second style and updated function calls as needed - in next commit.

lib/blacklight_range_limit/controller_override.rb
peetucket: rolled back in next commit to old style

app/assets/javascripts/blacklight_range_limit/range_limit_shared.js
peetucket: Also, made parseNum method more general - it simply tosses anything that is not a digit, so is agnostic to commas, periods, etc. as separators.

app/assets/javascripts/blacklight_range_limit/range_limit_shared.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  BlacklightRangeLimit.noConflict = function noConflict() {
+    global.BlacklightRangeLimit = previousBlacklightRangeLimit;
+    return BlacklightRangeLimit;
+  };
+
+  BlacklightRangeLimit.parseNum = function parseNum(str) {
+    str = String(str).replace(/[^0-9]/g, '');
+    return parseInt(str);
```
cbeer: @mejackreed should this be parseInt(str, 10)? I seem to remember something funny like that popping up somewhere else (and can't remember why..)

app/assets/javascripts/blacklight_range_limit/range_limit_distro_facets.js
```diff
@@ -204,8 +204,8 @@ Blacklight.onLoad(function() {
           segment = find_segment_for(pos.x);
 
           if(segment != last_segment) {
-            $('.distribution').tooltip('destroy');
-            $('.distribution').tooltip({'title': function() { return find_segment_for(pos.x).label  + ' (' + segment.count + ')' }, 'placement': 'bottom', 'trigger': 'manual', 'delay': { show: 0, hide: 100}});
+            $('.distribution').tooltip('destroy');
```
cbeer: Can you remove this extra whitespace?

app/assets/javascripts/blacklight_range_limit/range_limit_shared.js
mejackreed: Ahh.. yep good catch @cbeer some unexpected behavior can happen:

http://fczaja.blogspot.com/2010/09/javascript-parseint-gotcha.html

mejackreed:
:+1:

PR #37, opened 2015-08-26
Requires BL 5.15+ and runs without deprecation

jrochkind:
I followed cbeer's lead
projectblacklight/blacklight_advanced_search@ca1b13acb165d05
for some integration. That is, a module included in CatalogController,
on inclusion will include a module in the configured SearchBuilder.
This is a bit wacky, but I wasn't sure the right safe way to do this
for all the current configuration possibilities, so just followed
cbeer's lead.

Note also at
https://github.com/projectblacklight/blacklight_range_limit/blob/bl514/lib/blacklight_range_limit/controller_override.rb#L41-L45
some contortions needed to execute a search_results intended to be
"default processor chain, but with one method removed, and another one
added", and do it in a safe way for various configurations. This may
be something BL API should support better built-in, with arguments
to SearchBuilder/CatalogController#search_results ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,16 @@ sudo: false
 rvm:
   - 2.2.2
 
+before_install:
+  - gem install bundler
+
+env:
+ - "RAILS_VERSION=4.2.3"
+
 notifications:
   irc: "irc.freenode.org#blacklight"
   email:
       - blacklight-commits@googlegroups.com
+
+global_env:
+  - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,7 @@ language: ruby
 cache: bundler
 sudo: false
 rvm:
-  - 1.9.3
-  - 2.0.0
+  - 2.2.2
 
 notifications:
   irc: "irc.freenode.org#blacklight"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,12 @@ notifications:
   email: false
 
 language: ruby
-cache: bundler
 sudo: false
 rvm:
   - 2.2.2
 
 before_install:
   - gem install bundler
-
-env:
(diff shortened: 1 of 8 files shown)


jrochkind:
I can not get tests to run locally. When I try `rake spec`, I just get a terminal that hangs forever. 

Any advice or guidance?  I need to run tests locally to get tests to pass.

jrochkind:
I can't run master branch tests locally either, even off master I get:

     ActionView::Template::Error:
       undefined method `type' for .focus:Sass::Selector::Class
         (in /Users/jrochkind/code/blacklight_range_limit/spec/internal/app/assets/stylesheets/blacklight.css.scss)

Very frustrating.

jrochkind:
Okay, now `bundle exec rake` passes for me locally, but on travis:

     /home/travis/.rvm/rubies/ruby-2.2.2/lib/ruby/site_ruby/2.2.0/rubygems/dependency.rb:315:in `to_specs': Could not find 'railties' (= 4.2.4) among 16 total gem(s) (Gem::LoadError)

Not being able to reproduce locally,not quite sure what to do here. Hm.

jrochkind:
Okay, tests passing.I'm afraid I was kind of just bit twiddling to get it to work, me and engine_cart are not friends. 

* Had to _disable_ bundler caching on travis, this has something to do with engine_cart, cant' explain what
* Had to hard-code a `gem 'sass', "~> 3.4"` to Gemfile, not guarded by any condition. Otherwise it was using an old broken version of sass. Can't explain this one and I know it isn't totally right, but I got it to work.

jrochkind:
Updated to require BL 5.15, and use new feature to allow cleaner SearchBuilder per-call configuration in #range_limit

jrochkind:
Anyone want to code review? @jcoyne, @cbeer? 

If I don't hear anything, and nobody tells me not to, then I may merge and release myself next week. (Feel free to tell me not to if you want me to wait longer for code review).

PR #54, opened 2016-10-25
Avoid circular assignment reference

### LINKED ISSUE #53: Legacy Bugfix - circular argument reference
In the ancient 2.2.0 release and a modern ruby, we get warnings

```
/Users/atz/.rvm/gems/ruby-2.3.1/gems/blacklight_range_limit-2.2.0/lib/blacklight_range_limit/view_helper_override.rb:6: warning: circular argument reference - params
/Users/atz/.rvm/gems/ruby-2.3.1/gems/blacklight_range_limit-2.2.0/lib/blacklight_range_limit/view_helper_override.rb:19: warning: circular argument reference - params
```

These are easy to fix and would be nice to get a 2.2.1 release that resolves them.  The code is here, just need to change the variable name not to assign to "itself" circularly:
https://github.com/projectblacklight/blacklight_range_limit/blob/v2.2.0/lib/blacklight_range_limit/view_helper_override.rb#L6-L21

atz:
Newer rubies warn on this usage, and it actually breaks blacklight applications.

Fixes #53


atz:
Travis build is already broken on this branch, of course.

atz:
The only errors on rubies we still care about are routing, which I pretty clearly did not touch here.  Willing to merge and release anyway, since #c581b88 simply fixes my problems downstream.  Everything else is runaround due to decrepit age.

cbeer:
Looks like an error running the engine cart generator:

https://travis-ci.org/projectblacklight/blacklight_range_limit/jobs/170334815#L414

atz:
@cbeer I agree, though I don't really care to undertake fixing every outstanding problem just to rename a variable.

cbeer:
I understand that, although I suspect the modern ruby failures you're seeing in travis may be related to the engine_cart upgrade you've also made in this branch.

atz:
The actual file vs. what Engine Cart appears to be looking for:
```
/home/travis/build/projectblacklight/blacklight_range_limit/spec/test_app_templates/lib/tasks/blacklight_test_app.rake
/home/travis/build/projectblacklight/blacklight_range_limit/test_app_templates/lib/tasks/blacklight_test_app.rake
```

Which is to say, it has lost the `spec` context.

cbeer:
Looks like; and here's a commit in `master` that did the engine cart upgrade: https://github.com/projectblacklight/blacklight_range_limit/commit/cb53b8a0c08c68b1ce2242a8fb5f30df13b6d57a#diff-8a68e9cd749c03f8bc8179442775be6eL4

atz:
Is there some reason they back out to root and the re-enter spec instead of just not backing out of spec in the first place?

cbeer:
You aren't in `spec` at that point. With the older engine_cart, the test app was generated into `./spec/internal`, but we switched to `./.internal_test_app` at some point instead.

atz:
@cbeer: you have recommended adjusting our app to pull this gem from `circular_reference` branch rather than merge and release.  This seems unlikely to be as durable, given that an old topical branch would probably be deleted.  Even the target *release* branch was deleted before I re-pushed it this week.  I'm ok pulling from github if we can merge this and otherwise protect the branch from deletion.

atz:
@cbeer or anybody else: ^^ opinion on merge?

jcoyne:
There's still an error regarding a missing rake task when generating the application. Can you fix that?

atz:
@jcoyne: that error predates my changes, so no, I don't feel compelled to fix that or a host of other potential issues just to rename one variable.  You're not contending that the variable rename *caused* the error, right?  And that rename was both necessary and sufficient to fix my application's problem downstream.  It's now building against the topic branch.

jcoyne:
@atz No, i'm not saying you caused the error, but part of maintaining a project is ensuring a build is green.  If it's not building, we can't add any changes.

atz:
You can reason about whether the changes could *possibly* induce the build issues and reason that they provide *actual* value downstream and proceed accordingly.  Or you could block, and I'll stay using the topic branch (meaning it can never be deleted).  Or I'll decide to merge it unilaterally since you obviously don't care about consumers of the release branch, while I *am* a consumer of this branch.

Your build is broken today.  It was broken when I started.  I put in work towards fixing it, but I'm not adopting this gem, I'm fixing my app (and any other sad old consumers trying to migrate legacy code to new versions of ruby).

jcoyne:
@atz as you are presumably working on this for Stanford and something like 90% of the commits in the past year are from Stanford employees, _WE_ need to maintain this project.  If the build is broken, it is Stanford's problem.  If this project is not well maintained, then any software you are using it for is not going to have high quality.  The next developer who has to do an upgrade is  going to run into the same difficulty the next time that upgrades are needed.  Nobody else is going to fix this for us.  If you are feeling stuck, please ask one of the other developers for help.

atz:
@jcoyne, how 'bout you fix the broken build in master (#56), and I'll work on it for `2.x`?

atz:
@jcoyne Well, having fixed master, consider this to be me asking you for help w/ `release-2.x`. It turns out I had already updated engine_cart and its generated stanza here in this branch.

jrochkind:
@atz I just did a javsacript-only change off master that passed CI. Master seems to pass CI now. Will this pass CI if you rebuild it? Sorry, a bit confused as to current state. 

I'd like to do a bugfix release of blacklight_range_limit, can we actually get this in there easily after all, just by rebuilding?

lib/blacklight_range_limit/view_helper_override.rb
```diff
@@ -16,10 +16,10 @@ def has_search_parameters?
       super || has_range_limit_parameters?
     end
 
-    def query_has_constraints?(params = params)         
-      super || has_range_limit_parameters?(params)
+    def query_has_constraints?(my_params = params)
+      super(my_params) || has_range_limit_parameters?(my_params)
```
jrochkind: `super` called without args means "pass the exact same args that were passed in to me through to super", so the change from bare `super` to one with args was not neccesary -- the previous one would have kept working despite parameter name change. 

Not trying to block this merge, just commenting for future knowledge!

atz:
@jcoyne Stanford has far more code than it is my personal responsibility to redeem, particularly along legacy branches/releases.  I will make an additional good faith effort, but I see that you have not responded in 5 months.  At some point, I will decide that I am the only person that *actually* cares about this branch and merge accordingly.

atz:
Current dependency failure is from `mime-types-data`, which according to its gemspec's history, always required ruby 2.0.  So I need to back down `mime-types` to some point before it introduced `mime-types-data`.

atz:
Even though my gemspec now has `"nokogiri", "< 1.7.0"` (and the gem itself finally installs OK), the `engine_cart` build dies on requesting a too-new nokogiri:  
```
Gem::InstallError: nokogiri requires Ruby version >= 2.1.0.
An error occurred while installing nokogiri (1.7.1), and Bundler cannot continue.
```

atz:
Tests now run, despite the same engine_cart app's failure to install nokogiri.  Tests fail in all rubies with `ActionController::RoutingError: No route matches [GET] "/catalog"`

jrochkind:
@atz I can't figure out why tests are failing for you, when I just had a PR with tests succeeding. So it's not a legacy issue -- master passes. 

You changed an awful lot in here other than just the "avoid circular assignment references" as it says in the PR title. 

Maybe try to just commit the actual 'avoid circular assignment references' commit in a PR of it's own?  And separate the other stuff into other PR's if it's important?

I just did a PR that passed CI, CI is passing on master. If it's failing for this PR and not on master, it's gotta be _something_ in here that makes it fail, right?  But this PR includes a whole lot more than the label on the tin says it will.

jrochkind:
(I am a committer here, and wrote ~85%+ of the code in here too, and i'm working on this stuff again, for now -- I'm happy to merge a PR for you if it gets in a sensible state, and to try to help you get CI to pass, but with 15 commits in this PR only one of which is actually about "Avoid circular assignment reference", I gotta admin I don't really know what's going on, or why it's failing CI. I'd rather merge PR's that each do one thing than an omnibus PR, even with passing tests).

atz:
@jrochkind because this PR is not against `master`, it is against an ancient release branch.

atz:
@jrochkind I changed a lot in here because it was *demanded* of me that I get this release branch "fully maintained", despite nobody else (including me) having an actual interest in doing that.  I only want something that allows me to upgrade an old app on 2.x (and ultimately get onto newer stuff). 

We have often avoided specifying upper bounds on deps, optimistically.  In practice, that means our gems don't build within a couple years *from the same gemspec*.  This is one such case.  

I would happily cherry-pick and merge my one commit against this old release branch, though quite obviously none of the tests would pass, because they do not pass for the release branch anyway.  However great things are on master branch does nothing to provide a migration path here.

jrochkind:
@atz I'm confused what do you mean by 'the release branch'?  Tests are passing on master -- but maybe not with more recent ruby versions that you added in travis, is that it?

I'm a committer here too, I'm not interested in 'demanding' anything of you, I can't speak for anyone else, but I can put some time aside to help you get what you actually need into there, but for me, it would require breaking it into individual issues. 

I agree with you there's no reason to lump it all together and fix everything at once in one PR, or make fixing one thing a pre-requisite for fixing another thing that can be fixed independently.  In fact, trying to fix everything at once can in my opinion make things much more difficult, jumbling them all together and making it hard to focus on the nature of the problem and the best solution to it.

jrochkind:
I see this is a pull to `release-2.x`, so that's "the release branch".  Can anyone explain what this branch is for, as opposed to master? The current blacklight_range_limit release is 6.1.1. You're trying to back-port a fix to blacklight_range_limit 2.x, do I have that right? Sorry if it was here in the comments and I missed it, but can you explain your use case for needing to do that?

In general, it's definitely hard for us to maintain very old branches, we don't even really have enough resources to maintain the _current_ branches, right?

atz:
Yes, the context here is essentially backport.  I'm trying to provide a 2.x release that allows an upgrade path rather than a dead-end trap.  Obviously, we have applications still on 2.x that need to be updated.  Here I'm not interested in *any* of the newer functionality.

Release branches are typically for limited activity: backporting bugfixes, security patches and maintenance releases (read: not feature development).  If we don't do at least those last two things, we're basically untrustworthy as a gem producer.  Per SemVer, the main thing is to respect the install context for that major version.  I.E., not asking consumers to upgrade ruby for a bugfix, or because our dependency specification has allowed components underneath to drift into incompatible terrain, as here.  

I'm not trying to make a 2.x release approaching master, I'm trying to make one that doesn't immediately result in an unbuildable application.  Otherwise, when I am trying to do `bundle update x y z` at some level of hydra stack, and a `~> 2.x` reference to this gem is swept up underneath one of those components, I am guaranteed to fail. Without its working lockfile, I couldn't even build the same app fresh.

jrochkind:
In general, I agree with you that release-maintenance has not historically been acceptable for things in the blacklight ecosystem.  I'm not sure what to do about it, I think a solution will probably neccesarily require _process_ changes, where there are actually known people with the authority and responsibility for a gem (or at least  it's release maintanance), instead of being whoever shows up at a given time. But. 

> Release branches are typically for limited activity: backporting bugfixes, security patches and maintenance releases (read: not feature development). If we don't do at least those last two things, we're basically untrustworthy as a gem producer.

The last 2.x release was August 2014.  (May have also been the last one I personally was involved with prior to now).  The gem is now up to 6.1.  I don't believe the only way to be trustworthy as a gem producer is to agree to maintain old versions _forever_, hardly anything in railsland does that, a 2-year-old and 4-major-version-old release would not be expected to be still maintained hardly anywhere. 

But I agree with you that blacklight ecosystem has not been trustworthy as a gem producer historically.  Not really sure how to fix it, honestly. 

But meanwhile, back to this gem and where we are not where we'd like to be -- I would like to help you meet your needs, and can make some time to do so, but I'm still not sure what's going on. Can you make a narrative list of the different things preventing the release branch from being 'good', whether that's from passing on CI or from being usable in your app. 

If it's _just_ a CI issue, and really all you need is c581b880 to meet your local needs -- should we consider releaseing a 2.2.x patch fix release, even without it passing on CI, and call it a day? Can we be reasonably confident we're not introducing regressions even failing in CI?  For instance, do you want to try out a c581b880 version in your app, and ensure it works as expected manually? This is a small change, if it appears to work, it probably works -- and if not, more bugfix releases can always be made, since apparently current 2.x release is unusable anyway? (I don't totallly understand this; just un-usable on a modern MRI version, is the issue?)

Alternately, if the issue for you is just 'upgrade path', would it make sense to simply remove blacklight_range_limit from your app on your local upgrade branch, upgrade the app, then add it back in on a more recent version?  I realize this is not ideal and evidence of poor release management practices, but it may be the most efficient thing you can do for your local needs anyway?  blacklight_range_limit integration is fairly simple, we're not talking about removing/commenting out huge swaths of code or anything for you to temporarily remove it and then put it back when you're in a position to use a more recent version.

atz:
@jrochkind I can't remove it and re-add it without manually preserving the relevant portions of my lockfile.  That involves internalizing all the intelligence of bundler, so isn't really viable.  

One problem is that the non-SemVer dependencies underneath the gem have drifted far enough away from the supported environment of our major version.  It happens.  The appropriate thing for that major version to do is to account for those incompatibilities by limiting the range of dependencies to those that work.  Obviously, with actively developed dependencies, we don't expect compatibility to last forever across all future versions of those deps (despite our gemspec saying just that!): over a long enough time, that window closes. Alternatively, we could just declare our version unsupported and ignore the problem, which is what we've done historically.  

Unfortunately (or fortunately?), in the intervening 6 months, I've been reassigned to other applications and range_limit was removed from the stack completely.  I did test it with just that change locally.  (It didn't fix all the app's issues, but it was an improvement.)  I am fine with https://github.com/projectblacklight/blacklight_range_limit/pull/54/commits/c581b88099648d48bd4c7a28a25e3a06705c9589 and no CI, since we can reason about the changes in obvious ways.

If we want CI, we'll need to get into all the dependency mgmt issues that followed.

jrochkind:
I don't really know the decision-making process here, but I personally would be fine with another PR that is just c581b88, only to a release branch, and with cutting a 2.x.x release off of that even with CI failing. After all, CI is failing on 2.x anyway, and 2.x is apparently pretty unusable already anyway,it won't effect master or any future 6.x+ versions,  it doesn't seem like a step backward to cut a 2.x.x with c581b88 -- if it solves a problem for you, I don't see any reason not to do it. 

I would put my mark of approval on it, and do the merge and release if it's appropriate. Is anyone else around in a position and desire to object to this?

On the other hand, if you no longer actually have a need for getting 2.x working better, maybe we don't need to do any of this, and just close the PR?

mejackreed:
I don't have any objection to releasing failing older versions. I think we should be realistic about the need and the amount of time spent needing to support this versions.

Per @atz comment, is the release even necessary anymore?

atz:
Whatever, just 15 commits, right?  Anyway, let's bury this.

PR #61, opened 2017-04-06
Re-use tooltip instead of destroy/create

jrochkind:
Previous destroy/create was making the tooltip change not work at all for
me in Chrome, with this error in js console:

 ~~~ tooltip.self-05afb17….js?body=1:236 Uncaught TypeError: Cannot read
    property 'trigger' of null
       at HTMLDivElement.complete (tooltip.self-05afb17….js?body=1:236)
       at HTMLDivElement.fn (jquery.self-bd7ddd3….js?body=1:4826)
       at HTMLDivElement.handle (transition.self-6ad2488….js?body=1:55)
       at HTMLDivElement.dispatch (jquery.self-bd7ddd3….js?body=1:5227)
       at HTMLDivElement.elemData.handle (jquery.self-bd7ddd3….js?body=1:4879)
       at Object.trigger (jquery.self-bd7ddd3….js?body=1:5131)
       at HTMLDivElement.<anonymous> (jquery.self-bd7ddd3….js?body=1:5861)
       at Function.each (jquery.self-bd7ddd3….js?body=1:371)
       at jQuery.fn.init.each (jquery.self-bd7ddd3….js?body=1:138)
       at jQuery.fn.init.trigger (jquery.self-bd7ddd3….js?body=1:5860)
~~~

I didn't quite get to the bottom of the bug -- I think it was maybe
trying to add a new tooltip before the first one was actually fully
removed, or something like that.

Was not able to get it working with destroy/new tooltip design. 

Switching to-reusing one tooltip object but changing it's title works
properly again.

Changing title of an existing Bootstrap tooltip is not really documented,
but seems to be established and known to do it this way, and is using BS
tooltip public methods to do it. Eg
http://stackoverflow.com/questions/9501921/change-twitter-bootstrap-tooltip-content-on-click

### DIFF
--- a/app/assets/javascripts/blacklight_range_limit/range_limit_distro_facets.js
+++ b/app/assets/javascripts/blacklight_range_limit/range_limit_distro_facets.js
@@ -2,12 +2,12 @@
 //= require blacklight/core
 
 /* A custom event "plotDrawn.blacklight.rangeLimit" will be sent when flot plot
-   is (re-)drawn on screen possibly with a new size. target of event will be the DOM element 
+   is (re-)drawn on screen possibly with a new size. target of event will be the DOM element
    containing the plot.  Used to resize slider to match. */
 
 Blacklight.onLoad(function() {
   // ratio of width to height for desired display, multiply width by this ratio
-  // to get height. hard-coded in for now. 
+  // to get height. hard-coded in for now.
   var display_ratio = 1/(1.618 * 2); // half a golden rectangle, why not
   var redrawnEvent = "plotDrawn.blacklight.rangeLimit";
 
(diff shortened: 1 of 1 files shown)


jrochkind:
@cbeer, @jcoyne, you are the next in line after me in number of commits to this gem, with 36 and 13 commits respectively. 

Would you like to review this before I merge myself, and release a bugfix version?  If not, or I don't hear from you otherwise, I will do so!

mejackreed:
👍 
looks great! This fixes some of the issues that I saw from earlier.

jrochkind:
Awesome. You think it's okay if I cut a release to rubygems with the bugfix?

PR #63, opened 2017-05-26
allow range limit and advanced search to install without conflicts

tampakis:
This change allows the blacklight_range_limit plugin to properly install when the application already has an existing search history and saved searches controller, adding the range limit behavior to the controllers rather than overwriting them. The advanced search plugin also modifies these controllers so this change makes it easier to install a blacklight application that uses both plugins.

### DIFF
--- a/lib/generators/blacklight_range_limit/install_generator.rb
+++ b/lib/generators/blacklight_range_limit/install_generator.rb
@@ -16,7 +16,7 @@ module BlacklightRangeLimit
 
     def install_search_builder
       path = 'app/models/search_builder.rb'
-      if File.exists? path
+      if File.exist? path
         inject_into_file path, after: /include Blacklight::Solr::SearchBuilderBehavior.*$/ do
           "\n  include BlacklightRangeLimit::RangeLimitBuilder\n"
         end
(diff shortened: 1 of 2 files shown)


lib/generators/blacklight_range_limit/install_generator.rb
```diff
@@ -26,7 +26,27 @@ def install_search_builder
     end
 
     def install_search_history_controller
-     copy_file "search_history_controller.rb", "app/controllers/search_history_controller.rb"
+      path = 'app/controllers/search_history_controller.rb'
+      if File.exist? path
```
jcoyne: How about adding a comment in here about why you're checking for the existence of this path.

jrochkind:
status of this one? Could be merged? Waiting for changes required by @jcoyne ? What should we do with it?

jrochkind:
@tampakis can you check in and let us know your understanding of the status of this PR? Thanks!

lib/generators/blacklight_range_limit/install_generator.rb
jrochkind: @jcoyne I believe the answer is just "because not every Blacklight app that may want blacklight_range_limit may be using search history functionality or have a SearchHistoryController". 

I arrive at this belief because I myself currently have such an app that is using blacklight_range_limit but not does not have a `app/controllers/search_history_controller.rb` :) 

@jcoyne if i make a commit `# because not every app has a app/controllers/search_history_controller.rb`, does that meet your needs?

jrochkind:
comment was already there, we were just all looking at an outdated diff github was showing us, oops!

PR #65, opened 2017-07-05
adds compatibility for rails 5.1, closes #64

### LINKED ISSUE #64: Incompatibility with rails 5.1
ActionController parameters no longer inherit from the Hash class in Rails 5.1 and the`any?` method no longer is defined in the [query_has_constraints? method](https://github.com/projectblacklight/blacklight_range_limit/blob/3e09cea8b1eae4d3401ef7bc5a45b0a17735abbc/lib/blacklight_range_limit/view_helper_override.rb#L15).

tampakis:
Converts the actioncontroller parameter object into a hash calling the `any?` method. Adds compatibility for rails 5.1, but requires at least rails version [4.2](http://api.rubyonrails.org/v4.2.0/classes/ActionController/Parameters.html#method-i-to_unsafe_h). 
Updated minimum supported Blacklight version to 6.0.2 raises the [minimum required rails version](https://github.com/projectblacklight/blacklight/blob/v6.0.2/blacklight.gemspec#L25) to 4.2.

### DIFF
--- a/blacklight_range_limit.gemspec
+++ b/blacklight_range_limit.gemspec
@@ -19,8 +19,8 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'rails', '>= 3.0'
   s.add_dependency 'jquery-rails' # our JS needs jquery_rails
-  s.add_dependency 'blacklight', '~> 6.0'
-  
+  s.add_dependency 'blacklight', '>= 6.0.2'
+
   s.add_development_dependency 'rspec', '~> 3.0'
   s.add_development_dependency 'rspec-rails'
   s.add_development_dependency 'capybara'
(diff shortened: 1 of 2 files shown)


blacklight_range_limit.gemspec
```diff
@@ -19,8 +19,8 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'rails', '>= 3.0'
   s.add_dependency 'jquery-rails' # our JS needs jquery_rails
-  s.add_dependency 'blacklight', '~> 6.0'
-  
+  s.add_dependency 'blacklight', '>= 6.0.2'
```
jcoyne: Is this going to work with Blacklight 7?

PR #68, opened 2017-08-29
adds the ability to configure form classes via engine config

mejackreed:
This PR adds the ability to configure form and submit button classes using engine config.


jcoyne:
I'm going to have to say 👎  on this one.  We already did something like this for schema.org stuff in Hyrax and we had to back it out because we'd end up copying the same config to every i18n file.   Can you make this a separate configuration file instead?

jcoyne:
@mejackreed thanks for changing to the Engine config. 👍

PR #71, opened 2017-11-30
Add the test:server rake task.

dkinzer:
In order to avoid having to rebuild the test server every time you run
a test locally a `test:server` task is added to the Rakefile that can
set the testing environment as expected for quicker iteration on a test
while working on it locally.

Thanks to @tampakis who posted this solution on the #blackligt code4lib
slack channel.


tampakis:
Thanks @dkinzer!

PR #75, opened 2017-12-20
Slider change should update text input values.

mohideen:
Currently, the slider change updates the text input values when `segments: true` is configured. This is because the event handler is added by segment initialization turnIntoPlot()->areaChart() at [L245](https://github.com/projectblacklight/blacklight_range_limit/blob/master/app/assets/javascripts/blacklight_range_limit/range_limit_distro_facets.js#L245).

This PR will allow the updates to happen even if `segments: false` is configured.


tampakis:
@mohideen, thanks for the contribution!

PR #81, opened 2018-04-29
Replace solr_field with field_name.

dkinzer:
When upgrading a blacklight site from BL-6.14.1 to BL-7 I got the
following error from this gem:

```
undefined local variable or method solr_field
``

This is because solr_field has been removed from blacklight:
https://github.com/projectblacklight/blacklight/pull/1621

This commit replaces the calls to solr_field that caused the error
I was seeing with field_name.


mejackreed:
Checkout this branch `blacklight-7` where I think a lot of these changes should be working: https://github.com/projectblacklight/blacklight_range_limit/tree/blacklight-7

We haven't shipped this to `master` yet as I think we were waiting for a release.

dkinzer:
@mejackreed sweet, that did the trick; closing this PR.

PR #90, opened 2018-09-19
Fix range facet fails in advanced form.

dkinzer:
When in advanced form the form selector does not work because the range
is not rendered as a form within a form.  Instead it is a div.

This updates the selector to try again if form selection failed the
first time around with a selector that works for the facet when it is
inside of the advanced form.


dkinzer:
Turns out trying to render facets within the form as I was doing is not a good idea because it messes with search in an unexpected way.

Rendering outside of the form poses no issues and this change is not required when the facet is rendered outside of the form.

PR #91, opened 2018-11-08
 Add the workaround for Hyrax-based applications

boveus:
Hey there folks,

I was struggling to get this gem implemented into our Hyrax-based application, but thanks to the Samvera community I was able to get up and running with a workaround.  I think it would be helpful to other folks trying to get this gem up and running to have a note about getting things to work with Hyrax in the Readme.

Thanks for all of your hard work on this gem.

- Brandon

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ is probably the best touch UI anyway, if it can be made to work well.
 
 ## Integrating with Hyrax
 
-If you are using an application based on the [Hyrax](https://github.com/samvera/hyrax) engine, there is some additional customization required to get this gem up and running.  You will need to create a new initializer within 
+If you are using an application based on the [Hyrax](https://github.com/samvera/hyrax) engine, there is some additional customization that is needed to get the range limit code into the proper 'SearchBuilder' class to get this gem running.  You can create a new initializer within 
 `config/initializers/` that injects the `BlackLightRangeLimit::RangeLimitBuilder` into your CatalogController. 
 ```ruby
 Rails.application.config.to_prepare do
@@ -146,7 +146,7 @@ Rails.application.config.to_prepare do
 end
 ```
 
-This workaround was tested in Hyrax version 2.3.2, but may apply to earlier or later versions as well as related projects (such as Sufia or Hyku).
+This workaround was tested in Hyrax version 2.3.2, but may apply to earlier or later versions as well as related projects (such as Sufia or Hyku).  This is not neccesarily the most elegant solution, so if you have a suggestion for improving this workaround we encourage you to submit a PR and let us know! 
 
 # Tests
 
--- a/README.md
+++ b/README.md
@@ -135,9 +135,9 @@ ourselves yet.
 Also not sure how well the flot select UI works on a touch screen. The slider
 is probably the best touch UI anyway, if it can be made to work well. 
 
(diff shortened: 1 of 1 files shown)


jrochkind:
Can you just change "there is some additional customization required to get this gem up and running.", to something like "Some people have found some additional code is needed to get the range limit code into the proper 'SearchBuilder' class". 

I'm also curious if/why that monkey-patch is actually needed. There MAY be a better way, editing the proper SearchBuilder class directly. 

I just don't want it to seem like this README is suggesting this is the "right" way or actually makes sense at all (it really doesn't, there should be a better way without monkey-patching)... just that it's what some people have figured out, there may be a better way. 

Make sense?

jrochkind:
I wonder if @no-reply or other Hyrax experts can suggest a better way?

boveus:
That sounds good.  I will modify the wording to make it clear that this is _a_ workaround and not necessarily a best practice.

README.md
```diff
@@ -135,6 +135,19 @@ ourselves yet.
 Also not sure how well the flot select UI works on a touch screen. The slider
 is probably the best touch UI anyway, if it can be made to work well. 
 
+## Integrating with Hyrax
+
+If you are using an application based on the [Hyrax](https://github.com/samvera/hyrax) engine, there is some additional customization that is needed to get the range limit code into the proper 'SearchBuilder' class to get this gem running.  You can create a new initializer within
```
mejackreed: Could we make this even more generic?

I know this is relevant right now, but I'm concerned with the specifics of Hyrax and versions where this might not age well in the readme. Perhaps it could be more generic to encompass "other Blacklight engines that inject into a `SearchBuilder`" or something like that?

no-reply:
Rather than extending the existing class with a monkey patch, why not use inheritance and Blacklight's configuration?

```ruby
class RangeLimitCatalogSearchBuilder < Hyrax::CatalogSearchBuilder
  include, BlacklightRangeLimit::RangeLimitBuilder
end
```

README.md
jrochkind: I actually specifically asked for/suggested the hyrax version number, because in my experience this has changed several times over sufia/hyrax history -- I was worried about it aging poorly _without_ that, since the hyrax maintainers are not going to be maintaining this README. 

Arguably the instructions should go in Hyrax somewhere... but I think I agree with @boveus that this is where people are going to look for them.

jrochkind:
@boveus want to try out @no-reply's suggestion and see if it works for you? It does seem cleaner to me, as something to suggest to people.

boveus:
Thank you for the suggestion @no-reply, that seems a lot clearer.  I just tested it on my dev environment and it seems to work, so I updated the instructions to reflect your suggestion.

@jrochkind @mejackreed do we have a consensus on if we should keep the version information in there?

no-reply:
I'd suggest that the version information isn't relevant.

This extension approach seems to be the one designed for Blacklight, and documenting it here generically seems like a good idea. The connection to Hyrax seems entirely incidental, to me.

jcoyne:
@boveus if this is a Hyrax specific fix, then perhaps it should go in a Hyrax documentation rather than here.  This gem's only relationship to Hyrax is that both are Blacklight plugins.

PR #92, opened 2019-01-21
Update compatibility to blacklight 7+

cbeer:


jrochkind:
blacklight_range_limit 7 is already released and works (basically, with some bug(s) there are PR's for) with Blacklight 7. 

But this PR updates blacklight_range_limit's gemspec to _require_ Blacklight 7. Which it doesn't right now, it allows any version of blacklight. To extent the horse has left the stable, since blacklight_range_limit 7.0.0 has already been released, which probably only _works_ with BL7, but doesn't express it in the gemspec. 

But not too late to make it right?  This is currently red, if you rebase on master to green, I'll approve and merge it?

mejackreed:
@jrochkind I've done the rebase and pushed up

jrochkind:
@mejackreed I'll take that as an endorsement for the idea that blacklight_range_limit indeed ought to be specifying a BL 7.x requirement, and approve and merge this. Thanks!

PR #94, opened 2019-02-18
make <span class="facet-count">	stay on 1 line

xhero:
It seems that if facet-count is on multiple lines the length calculation includes all the whitespace and gets confused.


jrochkind:
I hadn't noticed this, and ended up spending time debugging it and implementing a solution myself, which was merged in #101. So closing this as a dup. 

Thanks for your work @xhero , sorry it got neglected for so long (really sorry, as it led to me spending quite a bit of time duplicating your work!)

PR #100, opened 2019-05-07
Remove generated 'require tether'

### LINKED ISSUE #97: `tether` requirement?
The blacklight_range_limit generator adds a `//= require tether` to your application.js. 

This appears to have been added in commit c9275b, as part of many other changes to "Update to Blacklight 7", with no further comments. (Not sure if the tether dependency actually pre-dates this, but was just moved around?). 

I can't find any use of `tether` in blacklight_range_limit JS, or other documentation of what it's for and why it's needed. 

Can anyone shed some light?

If we document what it's for and how it's used, we'll have some chance of recognizing if it _no longer becomes_ necessary in the future, or has another better alternative. Try to avoid the risk of it staying in the codebase forever that it's generated into local app, when not necessarily needed/optimal, because nobody remembers what it's for.

jrochkind:
tether.js is not actually relevant to any blacklight_range_limit JS.

Earlier pre-releases of Bootstrap 4 required tether for tooltips. But even before final Bootstrap 4.0, Bootstrap tooltips were no longer using tether, using 'popper.js' instead. https://github.com/twbs/bootstrap/pull/22444

So the generated require tether (and rail gem to provide tether as a sprockets asset) are irrelevant to blacklight_range_limit.

It is not necessary to replace with a "require 'popper'":
* A standard generated blacklight app will already have a "require 'popper'" in it's application.js https://github.com/projectblacklight/blacklight/blob/55e748dd3465b9e2781d4def04b912fcbfb1b231/lib/generators/blacklight/assets_generator.rb#L20
* The failure mode for not having popper is graceful -- tooltips simply won't work, and you will get a line in JS console "Bootstrap's tooltips require Popper.js"

Closes #97

### DIFF
--- a/blacklight_range_limit.gemspec
+++ b/blacklight_range_limit.gemspec
@@ -20,7 +20,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'rails', '>= 3.0'
   s.add_dependency 'jquery-rails' # our JS needs jquery_rails
   s.add_dependency 'blacklight'
-  s.add_dependency 'tether-rails'
 
   s.add_development_dependency 'rspec', '~> 3.0'
   s.add_development_dependency 'rspec-rails'
--- a/lib/blacklight_range_limit/engine.rb
+++ b/lib/blacklight_range_limit/engine.rb
@@ -2,7 +2,6 @@ require 'blacklight'
 require 'blacklight_range_limit'
 require 'rails'
 require 'jquery-rails'
-require 'tether-rails'
 
 module BlacklightRangeLimit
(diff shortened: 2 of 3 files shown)


jrochkind:
Failure seems to be unrelated. 

```
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    engine_cart (~> 1.0) was resolved to 1.2.0, which depends on
      bundler (~> 1.5)
  Current Bundler version:
    bundler (2.0.1)
```

I can try to figure out how to fix; should probably be in a separate PR?

PR #101, opened 2019-05-07
Avoid whitespace in facet-count span for 'missing' line

### LINKED ISSUE #98: "Unknown" line is styled funny
On Blacklight/blacklight_range_limit 7.  A stock generated app, with not much in it except adding blacklight_range_limit, then indexing some stuff to test. 

Screenshot 2019-05-06 17 16 33

See how "Unknown" goes vertical, and the number is out of bounds over in the main column? Anyone else encountered this?

Investigating, the number looks like this in the chrome dev tools inspector:

```html
<span class="facet-count" style="width: 38ch;">
                1,106
              </span>
```

That `style="width: 38ch;"` doing it -- if I manually remove that in the inspector, it displays properly. 

But I can't figure out what is adding that "width: 38ch".  It's not always 38ch necessarily, I think it can depend on screensize. It's not Rails view templates. It is apparently some other JS doing a calculation and then adding this `style` attribute. 

But I can't figure out what JS, or how to stop it. This is a stock generated BL/blacklight_range_limit app, so it shouldn't be any custom/third-party JS. It's something that's in a stock app. There don't seem to be any special "data-" attributes on the relevant `span` that would be asking something to do something weird like this. 

Anyone have any ideas?

jrochkind:
Whitespace triggers a bug in some odd Blacklight JS that tries to calculate display size of facet-counts.

Closes #98

Also reported to Blacklight at:
https://github.com/projectblacklight/blacklight/issues/2098

### DIFF
--- a/app/views/blacklight_range_limit/_range_limit_panel.html.erb
+++ b/app/views/blacklight_range_limit/_range_limit_panel.html.erb
@@ -79,9 +79,9 @@
               <span class="facet-label">
                 <%= link_to BlacklightRangeLimit.labels[:missing], add_range_missing(field_name) %>
               </span>
-              <span class="facet-count">
-                <%= number_with_delimiter(stats["missing"]) %>
-              </span>
+              <%# note important there be no whitespace inside facet-count to avoid
+                  bug in some versions of Blacklight (including 7.1.0.alpha) %>
+              <span class="facet-count"><%= number_with_delimiter(stats["missing"]) %></span>
             </li>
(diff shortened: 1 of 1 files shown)


jrochkind:
Needs to be rebased on #102

jrochkind:
This may be a dup of #94, but that one has a messy commit history and unrelated changes and is currently red, so how about this one?

jrochkind:
Thank you!

PR #102, opened 2019-05-07
have travis install bundler 1.x

jrochkind:
engine_cart 1.x has a hard dependency on bundler 1.x (rather than 2.x). Until we can upgrade to engine_cart 2.x, need to make sure bundler 1.x is being used by travis

Best would be upgrading to engine_cart 2.x, which is appropriately agnostic about which version of bundler you use. But I made some very tentative forays there, and decided it was non-trivial to do that, and I personally didn't have time for it at present. This gets us back to green so it's not a blocker for other work on blacklight_range_limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,10 @@ rvm:
   - 2.3.1
 
 before_install:
-  - gem install bundler
+  # engine_cart 1.x has a hard-dependency on bundler 1.x. Until
+  # we can upgrade to engine_cart 2.x, need to make sure we have a bundler 1.x.
+  # and it's the latest bundler, so it will be used.
+  - gem install -v "~> 1.0" bundler --no-document
 
 notifications:
   irc: "irc.freenode.org#blacklight"


jrochkind:
(Got an initial travis failure, clicking "rebuild" on travis on it made it go green. So we have something flaky in travis, but I don't believe it's related to this change, it's pre-existing and a separate thing to be dealt with if needed separately)

jrochkind:
I do not know how to update engine_cart. When I tried simply updating to 2.x, it failed, and I do not have time/interest to figure it out and debug it. 

Right now travis is not green, which is a blocker to any other PR being considered. Is there a reason we can't just merge this simple and safe PR now to get it green, so engine_cart upgrade becomes an orthogonal concern instead of a blocker on any other work being reviewed in parallel?

jrochkind:
Ah, I see you got engine_cart upgraded in another PR, thanks! @cjcolvar would you like to review and merge that one, and if so I'll close this one? (Or I can do that myself I guess maybe? Confirm?)

One way or another, I just want to remove the "travis broken" blocker keeping me from getting other PR's green!

jrochkind:
Closing in favor of #103, thanks!

PR #105, opened 2019-05-07
remove explicit sprockets 'require jquery'

### LINKED ISSUE #99: `require 'blacklight_range_limit'` requires JQuery, leading to double-require or incompat
The way to include blacklight_range_limit's JS in your app is to put:

    require 'blacklight_range_limit'

In your application.js. This is also what the blacklight_range_limit generator does. 

This will do a `require 'jquery'`, which will require jquery from the jquery-rails gem into your JS manifest.  The [blacklight_range_limit README actually notes](https://github.com/projectblacklight/blacklight_range_limit#javascript-dependencies) this, so it was intentional at some point (possibly by a long-past me), and possibly long-standing -- but it turns out this is very problematic. 

Firstly, if your app separately has a `require 'jquery'` in your `application.js` (as for instance a default Blacklight-generated app will) -- your compiled JS from sprockets will include _two_ copies of JQuery. Which at best is inflating your JS asset quite a bit, and at worst may cause odd behavior as JQuery isn't meant to be included twice. (Not totally sure if it does, but it _could_ be responsible for some animation weirdness, if some animations are running twice in parallel). 

So I suppose you could just remove `require 'jquery'` from your application.js, to have only the single require from JQuery-rails -- although this is still odd. 

But it gets worse. `require jquery` from [jquery-rails](https://github.com/rails/jquery-rails) will require the latest JQuery 1.x.  But meanwhile, a JQuery 2.x and 3.x are out. jquery-rails [packages latest versions of these too](https://github.com/rails/jquery-rails#installation), which you can `require jquery2` or `require jquery3` to get. 

If you try `require jQuery3` in your `application.js` -- blacklight_range_limit will still be requiring JQuery 1.x. Now you have a double-include of JQuery in your sprockets-compiled JS asset, but with _two different versions_. 

This _does_ cause problems, although JQuery will sometimes work just fine, it sometimes won't, at least/especially in JS that happens in blacklight_range_limit, or *after* the `require blacklight_range_limit` that has introduced the double-compile. I ran into such problems; I didn't debug them in detail, I stopped when I figured out that the double-Jquery-require was happening, and that eliminating it eliminated my problems. 

I think that blacklight_range_limit should probably _not_ `require jquery` in it's JS. It should count on the parent app to be requiring jQuery already. As generated Blacklight apps will have this, it should be fine. It's possible this is a backwards imcompat for some existing apps which don't require JQuery anywhere _but_ via blacklight_range_limit, and were working anyway -- they would stop. 

But i think this is just wrong, and has to be fixed. 

(I think maybe when I or someone else wrote the `require jQuery`, we misunderstood how sprockets `require` worked, and thought it would not double-require. But it does).

jrochkind:
blacklight_range_limit code _does_ need JQuery to work (for it's own code and it's dependencies like flot).

It was probably on that basis that an explicit 'require jquery' was included in blacklight_range_limits sprockets-processed file to be included in the host app. While the host app may likely _also_ have a 'require jquery' in it's application.js, the assumption was that sprokets is smart enough not to require it twice (which _may_ be true, unsure).

However, 'require jquery' from jquery-rails is actually the (quite old) latest JQuery 1.x. The explicit require here precludes the host app from choosing to use the newer jquery 2.x or 3.x (which should probably work fine). If the host app _does_ try to 'require jquery3' to get jquery 3.x, it leads to a really confusing and hard to debug failure mode where two versions of JQuery are loaded on top of each other, and interfere with each other in odd ways.

By removing the explicit require, we can leave it up to the host app which version of JQuery to load. We can also let the host app decide to load jQuery from some _other_ place than the jquery-rails gem, such as a CDN, or via webpacker/yarn/npm.

This does mean that if an existing app is not explicitly loading JQuery at all, and just counting on blacklight_range_limit to do it -- it would have to be changed to 'require jquery' in its application.js or otherwise load JQuery. This seems like a pretty unlikely situation though; among other things, the Blacklight generator adds a 'require jquery' to the host apps application.js already.

Closes #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,24 +6,24 @@ BlacklightRangeLimit:  integer range limiting and profiling for Blacklight appli
 
 # Description
 
-The BlacklightRangeLimit plugin provides a 'facet' or limit for integer fields, that lets the user enter range limits with a text box or a slider, and also provides area charts giving a sense of the distribution of values (with drill down). 
+The BlacklightRangeLimit plugin provides a 'facet' or limit for integer fields, that lets the user enter range limits with a text box or a slider, and also provides area charts giving a sense of the distribution of values (with drill down).
 
-The primary use case of this plugin is for 'year' data, but it should work for any integer field. It may not work right for negative numbers, however. 
+The primary use case of this plugin is for 'year' data, but it should work for any integer field. It may not work right for negative numbers, however.
 
-Decimal numbers and Dates are NOT supported; they theoretically could be in the future, although it gets tricky. 
+Decimal numbers and Dates are NOT supported; they theoretically could be in the future, although it gets tricky.
 
 
 # Requirements
 
-A Solr integer field. Depending on your data, it may or may not be advantageous to use a tint (trie with non-zero precision) type field. 
+A Solr integer field. Depending on your data, it may or may not be advantageous to use a tint (trie with non-zero precision) type field.
 
 ## Note on solr field types
 
-If all your integers are the same number of digits, you can use just about any solr type, including string/type, and all will be well. But if your integers vary in digits, strings won't sort correctly, making your numbers behave oddly in partitions and limits. This is also true if you use a pre-1.4 "integer"/pint/solr.IntField  field -- these are not "sortable". 
+If all your integers are the same number of digits, you can use just about any solr type, including string/type, and all will be well. But if your integers vary in digits, strings won't sort correctly, making your numbers behave oddly in partitions and limits. This is also true if you use a pre-1.4 "integer"/pint/solr.IntField  field -- these are not "sortable".
 
-You need to use a "sortable" numeric-type field. In Solr 1.4, the standard "int"/solr.TrieIntField should work fine and is probably prefered. For some distributions of data, it may be more efficient to use "tint" (solr.TrieIntField with non-zero precision). 
+You need to use a "sortable" numeric-type field. In Solr 1.4, the standard "int"/solr.TrieIntField should work fine and is probably prefered. For some distributions of data, it may be more efficient to use "tint" (solr.TrieIntField with non-zero precision).
 
-The pre Solr 1.4 now deprecated sint or slong types should work fine too. 
+The pre Solr 1.4 now deprecated sint or slong types should work fine too.
 
 # Installation
 
@@ -33,9 +33,9 @@ Add
 
     gem "blacklight_range_limit"
 
-to your Gemfile. Run "bundle install". 
+to your Gemfile. Run "bundle install".
 
-Then run 
+Then run
 
     rails generate blacklight_range_limit:install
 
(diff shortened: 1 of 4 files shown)


jrochkind:
Hmm, very odd and seemingly unrelated failure, can't figure out why we're getting it only now. 

```
Installing capybara 3.18.0
Gem::RuntimeRequirementNotMetError: capybara requires Ruby version >= 2.4.0. The
current ruby version is 2.3.1.112.
An error occurred while installing capybara (3.18.0), and Bundler
cannot continue.
Make sure that `gem install capybara -v '3.18.0' --source
'https://rubygems.org/'` succeeds before bundling.
In Gemfile:
  capybara
```

But MRI 2.3 is EOL as of 2019-03-31, I guess I'll do a separate PR to update ruby under test.

jrochkind:
Okay, still mysteriously failing. Travis gives an uninformative error message:

```
EngineCart failed on with: pid 7789 exit 1
/home/travis/.rvm/gems/ruby-2.5.5/gems/engine_cart-2.2.1/lib/engine_cart/tasks/engine_cart.rake:108:in `block (3 levels) in <top (required)>'
/home/travis/.rvm/gems/ruby-2.5.5/gems/engine_cart-2.2.1/lib/engine_cart/tasks/engine_cart.rake:139:in `block (2 levels) in within_test_app'
/home/travis/.rvm/gems/ruby-2.5.5/gems/bundler-2.0.1/lib/bundler.rb:313:in `block in with_clean_env'
/home/travis/.rvm/gems/ruby-2.5.5/gems/bundler-2.0.1/lib/bundler.rb:562:in `with_env'
/home/travis/.rvm/gems/ruby-2.5.5/gems/bundler-2.0.1/lib/bundler.rb:313:in `with_clean_env'
/home/travis/.rvm/gems/ruby-2.5.5/gems/engine_cart-2.2.1/lib/engine_cart/tasks/engine_cart.rake:138:in `block in within_test_app'
/home/travis/.rvm/gems/ruby-2.5.5/gems/engine_cart-2.2.1/lib/engine_cart/tasks/engine_cart.rake:137:in `chdir'
/home/travis/.rvm/gems/ruby-2.5.5/gems/engine_cart-2.2.1/lib/engine_cart/tasks/engine_cart.rake:137:in `within_test_app'
/home/travis/.rvm/gems/ruby-2.5.5/gems/engine_cart-2.2.1/lib/engine_cart/tasks/engine_cart.rake:104:in `block (2 levels) in <top (required)>'
/home/travis/.rvm/gems/ruby-2.5.5/gems/rake-12.3.2/exe/rake:27:in `<top (required)>'
/home/travis/.rvm/gems/ruby-2.5.5/bin/ruby_executable_hooks:24:in `eval'
/home/travis/.rvm/gems/ruby-2.5.5/bin/ruby_executable_hooks:24:in `<main>'
Tasks: TOP => default => ci => engine_cart:generate
(See full trace by running task with --trace)
```

If I run `bundle exec rake ci` locally, I get a more informative error message:

```
Bundler could not find compatible versions for gem "rails":
  In snapshot (Gemfile.lock):
    rails (= 5.0.2)

  In Gemfile:
    rails (~> 5.0.2)

    blacklight_range_limit was resolved to 7.0.0, which depends on
      blacklight was resolved to 7.1.0.alpha, which depends on
        rails (~> 5.1)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

But I don't understand what's going on. Apparently BL 7.1 requires Rails 5.1?  But then why isn't my app rails 5.1? Maybe I need to clean out my engine_cart test app. But I can't run `rake engine_cart:clean` without getting the same error. So on to figuring out how to do that manually, or maybe I'll do a fresh checkout. 

I wonder if the upgrade to engine_cart 2 is actually causing problems. Maybe we should have went with locking to bundler 1.x for now.

jrochkind:
OK, after `rm -rf Gemfile.lock .internal_test_app` and `bundle exec rake ci` I now have an error locally actually related to this PR, working on it.

jrochkind:
AND we're green!

jrochkind:
Thank you!

PR #107, opened 2019-05-08
Changing to the way other deselects are treated in Blacklight 7

cdmo:
rebased on master

### DIFF
--- a/app/views/blacklight_range_limit/_range_limit_panel.html.erb
+++ b/app/views/blacklight_range_limit/_range_limit_panel.html.erb
@@ -13,10 +13,10 @@
       <li class="selected">
         <span class="facet-label">
           <span class="selected"><%= range_display(field_name) %></span>
-           <%= link_to remove_range_param(field_name), :class=>"remove", :title => t('blacklight.range_limit.remove_limit') do %>
-           <span class="remove-icon">✖</span>
+          <%= link_to remove_range_param(field_name), :class=>"remove", :title => t('blacklight.range_limit.remove_limit') do %>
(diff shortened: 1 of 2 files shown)


cdmo:
Sorry for the mess!

app/views/blacklight_range_limit/_range_limit_panel.html.erb
```diff
@@ -14,7 +14,7 @@
         <span class="facet-label">
           <span class="selected"><%= range_display(field_name) %></span>
            <%= link_to remove_range_param(field_name), :class=>"remove", :title => t('blacklight.range_limit.remove_limit') do %>
-            <span class="glyphicon glyphicon-remove"></span>
+           <span class="remove-icon">✖</span>
```
mejackreed: Kind of minor, but could add some spacing since we have to touch this line. I didn't realize at first there was a block here.

cdmo: What do you think of that realignment? The `link_to` block is a sibling of the selected `span` so this alignment made sense to me. I also got rid of an extra space and some empty lines.

As an aside, not that it would catch all erb templating format problems with default settings, but [erb-lint](https://github.com/Shopify/erb-lint) is a nice gem for scanning erb template files for problems. We've been using it for our Blacklight work.

mejackreed: Nice thanks for that. Looks great.

jrochkind:
Thanks y'all!

PR #110, opened 2019-10-28
Fix travis ci broken.

dkinzer:


solr/sample_solr_documents.yml
```diff
@@ -1,2692 +0,0 @@
----
-- lc_1letter_facet:
```
mejackreed: Are removing these affecting the build? I found it sometimes helpful in the past to be able to have these handy.

dkinzer: I don't think so.  I did this because blacklight removed them from their repo, but I can add them back.

dkinzer:
@mejackreed I added that file back, but note that it does not get used by any of the rake tasks.  Both `blacklight:solr:seed` and `blacklight_range_limit:seed` use the files in `spec/fixtures`.  So users would have to know those files exist and ingest them manually.

mejackreed:
Thanks @dkinzer 👍

PR #115, opened 2019-11-26
Reorder components and enable larger view in modal of range content.

mejackreed:
blrl


cdmo:
This is cool, nice upgrade and cleanup! Thanks! I am curious, was there UX pre-work or something? I am wondering about that, sounds interesting.

mejackreed:
There is some stuff in here: https://github.com/sul-dlss/dlme/issues/694

And @ggeisler is on vacation, but has more comprehensive mocks and information about the UX side of things.

cdmo:
Awesome, thanks for sharing.

PR #122, opened 2019-12-04
Update range limit buttons/labels

jkeck:
* Use a button group
  * Move labels after inputs to support BS button groups
* Change default button label to Apply

Closes sul-dlss/dlme#764

## Before
Screen Shot 2019-12-04 at 11 52 12 AM

## After 
Screen Shot 2019-12-04 at 10 45 57 AM

### DIFF
--- a/app/helpers/range_limit_helper.rb
+++ b/app/helpers/range_limit_helper.rb
@@ -14,9 +14,9 @@ module RangeLimitHelper
 
     default = params["range"][solr_field][type] if params["range"] && params["range"][solr_field] && params["range"][solr_field][type]
 
-    html = label_tag("range[#{solr_field}][#{type}]", input_label, class: 'sr-only') if input_label.present?
-    html ||= ''.html_safe
-    html += text_field_tag("range[#{solr_field}][#{type}]", default, :maxlength=>maxlength, :class => "form-control range_#{type}")
+    html = text_field_tag("range[#{solr_field}][#{type}]", default, :maxlength=>maxlength, :class => "form-control text-center range_#{type}")
+    html += label_tag("range[#{solr_field}][#{type}]", input_label, class: 'sr-only') if input_label.present?
+    html
   end
 
   # type is 'min' or 'max'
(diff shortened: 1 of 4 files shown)


jrochkind:
Hi, to me this UI change to the text boxes seems confusing, it's less clear that you can edit them, and that they represent "From" and "to" dates. But that's just my opinion, may not be representative. 

Curious if you can provide any background about what motivated this change, was it the result of user testing or what? Wonder if it would be better to have a reviewer/approver from a separate institution before making UI changes?

ggeisler:
@jrochkind The change in input/submit button styling was part of the overall rearrangement of elements suggested by me and our other Stanford access team UX designer. We thought these changes would make it easier for users to quickly scan the date range facet elements and understand how they fit together. 

I agree that the new input/submit button styling loses a small amount of clarity; it is less immediately obvious the inputs are for entering "from" and "to" dates. We did discuss this before recommending the change. The motivation for making the change was that, for some viewport widths (at least when using the default 3 Bootstrap columns for the facet sidebar) the previous input/submit layout resulted in those elements wrapping to two rows, which increases the vertical height of the facet content area and reduces the cohesiveness of those elements:

Screen Shot 2019-12-18 at 11 32 36 AM

The new layout of these elements is a tradeoff, but does have the advantages of:

- The associated elements always display in a single row and are clearly associated with each other
- The height of the facet content area stays as compact as possible, making it easier for the user to scan all the elements quickly to understand what's going on

Screen Shot 2019-12-18 at 11 33 32 AM

And it seems likely that if the user is initially less clear that the inputs are for inputing "from" and "to" dates, that uncertainly will go away as soon as the user adjusts a date using the graph, slider, or one of the inputs. 

So I understand your concern, but like many UI decisions it's a tradeoff and we believe the change is an overall improvement and unlikely to cause real confusion. We haven't tested it, though (we don't have the resources for that right now), so if we see evidence in the future that this change causes more confusion that we thought, I certainly think we can reevaluate the tradeoffs.

PR #125, opened 2019-12-09
Add aria label to fix labeling accessibility issue

lfarrell:
Small fix to improve accessibility

### DIFF
--- a/app/assets/javascripts/blacklight_range_limit/range_limit_slider.js
+++ b/app/assets/javascripts/blacklight_range_limit/range_limit_slider.js
@@ -72,7 +72,7 @@ function isInt(n) {
       var begin_el = form.find("input.range_begin");
       var end_el = form.find("input.range_end");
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/blacklight_range_limit/range_limit_slider.js
```diff
@@ -72,7 +72,7 @@ function isInt(n) {
       var begin_el = form.find("input.range_begin");
       var end_el = form.find("input.range_end");
 
-      var placeholder_input = $('<input type="text" data-slider-placeholder="true" style="width:100%;">').appendTo(range_element);
+      var placeholder_input = $('<input type="text" aria-label="date slider" data-slider-placeholder="true" style="width:100%;">').appendTo(range_element);
```
jkeck: I wonder if this should just be a hidden input instead of text.  It's hidden via CSS anyway, and as a hidden input wouldn't require a label.

Is there a reason why this should continue to be a text input that I'm not noticing?

lfarrell: @jkeck I think you're right. I went ahead and updated the pull request.

PR #139, opened 2020-03-19
Guard against unconfigured range facets.

jkeck:


lib/blacklight_range_limit/segment_calculation.rb
```diff
@@ -15,6 +15,8 @@ def add_range_segments_to_solr!(solr_params, facet_field, min, max)
       raise InvalidRange, "The min date must be before the max date" if min > max
       field_config = blacklight_config.facet_fields[facet_field.to_s]
 
+      return solr_params unless field_config
```
jkeck: What we're ultimately trying to guard against is that `BlacklightRangeLimit.range_config` will return `false` in the case that it cannot find a configured facet.  Code later down the line expects hash access to this config.  I went down the path of changing that code, but am concerned about the API change possibly necessitating a major version release.

This change is in the spirit of the current API and should be able to be released in a patch version.

PR #143, opened 2020-04-01
Backport commits for a 6.4 release

jkeck:


blacklight_range_limit.gemspec
```diff
@@ -17,15 +17,19 @@ Gem::Specification.new do |s|
...
-  s.add_dependency 'rails', '>= 3.0'
+  s.add_dependency 'rails', '>= 4.2', '< 6'
   s.add_dependency 'jquery-rails' # our JS needs jquery_rails
-  s.add_dependency 'blacklight', '>= 6.0.2'
+  s.add_dependency 'blacklight', '>= 6.10.0', '< 7.0.0'
 
+  s.add_development_dependency 'chromedriver-helper'
   s.add_development_dependency 'rspec', '~> 3.0'
   s.add_development_dependency 'rspec-rails'
+  s.add_development_dependency 'sprockets', '< 4'
```
jkeck, line 27: I would prefer to not have to do this, but couldn't recall how to deal with the sprockets 4 issues (and seemingly other strategies didn't pan out).  Open for suggestions on how do handle this.

mejackreed: This thing maybe? https://github.com/geoblacklight/geoblacklight/blob/85ad59e79661e4c42f74249a88aff116ee5b0fb8/spec/test_app_templates/Gemfile.extra

jkeck: I'll give that a go!

jrochkind: (Not totally sure if a development_dependency will apply to my app when using a gem -- but I have an app using sprockets 4 with blacklight_range_limit already, so please don't do anything so that's not possible in a future release!)

jkeck: Yeah, totally agree @jrochkind.  I definitely want to avoid making this requirement for a consumer.

PR #148, opened 2020-04-07
Retreive chart values from a data attribute rather than the rendered …

mejackreed:
…text

This enables downstream consumers to customize the from/to text displayed.

### DIFF
--- a/config/locales/blacklight_range_limit.en.yml
+++ b/config/locales/blacklight_range_limit.en.yml
@@ -9,5 +9,5 @@ en:
       view_distribution: 'View distribution'
       view_larger: 'View larger »'
       results_range_html: 'Current results range from <span class="min">%{min}</span> to <span class="max">%{max}</span>'
-      single_html: '<span class="single">%{begin}</span>'
-      range_html: '<span class="from">%{begin}</span> to <span class="to">%{end}</span>'
+      single_html: '<span class="single" data-blrl-single="%{begin}">%{begin}</span>'
+      range_html: '<span class="from" data-blrl-begin="%{begin}">%{begin}</span> to <span class="to" data-blrl-end="%{end}">%{end}</span>'
--- a/config/locales/blacklight_range_limit.en.yml
+++ b/config/locales/blacklight_range_limit.en.yml
@@ -9,5 +9,5 @@ en:
       view_distribution: 'View distribution'
       view_larger: 'View larger »'
       results_range_html: 'Current results range from <span class="min">%{min}</span> to <span class="max">%{max}</span>'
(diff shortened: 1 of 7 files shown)


config/locales/blacklight_range_limit.en.yml
```diff
@@ -9,5 +9,5 @@ en:
       view_distribution: 'View distribution'
       view_larger: 'View larger »'
       results_range_html: 'Current results range from <span class="min">%{min}</span> to <span class="max">%{max}</span>'
-      single_html: '<span class="single">%{begin}</span>'
-      range_html: '<span class="from">%{begin}</span> to <span class="to">%{end}</span>'
+      single_html: '<span class="single" data-blrl-single="%{begin}">%{begin}</span>'
```
cbeer, line 12: Should we pass in a different value for the data attribute?

mejackreed: I'm not sure. What would expect? Seems like we are grabbing directly from params. Should we enhance that someway?

https://github.com/projectblacklight/blacklight_range_limit/blob/fbb307b27f86a3b023d864fcdab7f9ecc12ce762/app/helpers/range_limit_helper.rb#L39-L49

cbeer: I think I'm not understanding how one would change the displayed text if the value is passed through to both place.

mejackreed: I've done something in an incredibly hacky way here: https://github.com/sul-dlss/dlme/pull/902/files#diff-0d7881920a54eb448deb3f94e3fed85cR9-R26

The `from` text values are drawn directly from Solr -> read by JavaScript. However, I'm not sure if there is a better way to accomplish this?  Should we pass them through another translation here? https://github.com/projectblacklight/blacklight_range_limit/blob/fbb307b27f86a3b023d864fcdab7f9ecc12ce762/app/helpers/range_limit_helper.rb#L45

cbeer: Got it; that was the missing context. I think it'd provide a cleaner seam if we passed in the value for the attribute separate from the value for display (perhaps the value for display is pulled out into its own overridable helper method (or passed-in-lambda or otherwise) so we don't have to go parsing pre-formed HTML?)

mejackreed: Ok.. ill look into making that change.

mejackreed: I've made the change in db854ee91d301ef350d97ef887897bcf3a8a0346 and also adapted this change in https://github.com/sul-dlss/dlme/pull/902/commits/427e56906e0df1250096645d30e853a874aba6ab

PR #155, opened 2020-06-24
Do not return range_params for empty string inputs…

jkeck:
…(e.g. like what AdvancedSearch will send).


jrochkind:
Awesome, I had started having the problem I think you are fixing here in a recent update of blacklight and blacklight_range_limit dependencies, but hadn't had time to debug it yet. Thank you!

jkeck:
You're welcome.  Hopefully this solves it for the both of us (after I backport it. Unfortunately we're still on 6.x)

PR #158, opened 2020-06-26
Use the presenter to render values; fixes #157

### LINKED ISSUE #157: Calls deprecated facet_display_value
https://github.com/projectblacklight/blacklight_range_limit/blob/0b2e1a9059381ec6a7daad8a776e75b8dc8c0459/app/views/blacklight_range_limit/_range_segments.html.erb#L10-L14

`facet_display_value` is currently deprecated in Blacklight 7.9.0, says:

```
DEPRECATION WARNING: facet_display_value is deprecated and will be removed from blacklight 8.0. (called from facet_display_value at /Users/jrochkind/.gem/ruby/2.6.6/gems/blacklight-7.9.0/app/helpers/blacklight/facets_helper_behavior.rb:269)
```

I don't know what the non-deprecated alternative to `facet_display_value` is. Anyone know?

cbeer:


jrochkind:
Thanks! It's very helpful to see what BL maintainers consider the non-deprecated way to do this. 

It is already very hard to reliably keep track of what version of blacklight_range_limit goes with what version of BL.  I think I’d be more partial for putting a conditional in there to work with both BL versions, although of course that’s got some real downsides too.

Curious if anyone else has an opinion.

Personally, I still think “needing to format a value from solr for display” is something BL should support apps/plugins accessing through simple  API — that is, BL should continue to have a non-deprecated facet_display_value method, that HAS that implementation from #158 in there. But I think I understand that the Blacklight maintainers do not agree, so, okay, we’ll deal with it.

Curious for feedback from anyone else on any of this.

mejackreed:
I'm 👍 for using the Gemspec as the way to track/confirm our dependencies rather than dropping that into the code.

jrochkind:
just to clarify that means that anyone that wants to get any bugfixes or improvements in a future release of blacklight_range_limit will also be forced to update blacklight. For me, that would probably make me hold off on updating blacklight_range_limit unless I really had to, and had a bunch of time to set aside to deal with blacklight update, based on recent experiences with blacklight updates.

mejackreed:
Yes. If we want to utilize the non-deprecated methods here in range limit we should be ok with that tradeoff.

PR #169, opened 2021-03-30
Do not render constraint for un configured field.

### LINKED ISSUE #168: Range limit constraint does not take into account blacklight config
Our blacklight application has two separate blacklight range fields that are not part of the same configuration or even on the same controller.  However when we switch between the controller paths the application will render all the range limit param constraints even if that controller should not know about all the range limit fields that are part of the query parameters.

dkinzer:
Fixes #168

Fixes issue where a constraint gets rendered even though the controller
should not know anything about the particular field whose constraint is
being rendered.

### DIFF
--- a/lib/blacklight_range_limit/view_helper_override.rb
+++ b/lib/blacklight_range_limit/view_helper_override.rb
@@ -89,8 +89,8 @@
     def range_params(my_params = params)
       return {} unless my_params[:range].is_a?(ActionController::Parameters) || my_params[:range].is_a?(Hash)
 
-      my_params[:range].select do |_solr_field, range_options|
-        next unless range_options
+      my_params[:range].select do |solr_field, range_options|
+        next unless range_options && blacklight_config.facet_fields[solr_field].present?
 
         [range_options['missing'].presence,
          range_options['begin'].presence,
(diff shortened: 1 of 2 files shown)


lib/blacklight_range_limit/view_helper_override.rb
```diff
@@ -89,8 +89,10 @@ def range_config(solr_field)
     def range_params(my_params = params)
       return {} unless my_params[:range].is_a?(ActionController::Parameters) || my_params[:range].is_a?(Hash)
 
+      facet_field_keys = blacklight_config.facet_fields.keys
       my_params[:range].select do |_solr_field, range_options|
```
cbeer, line 93: By convention, now that `_solr_field` is used, we should drop the `_`. I also don't think we need to memoize the list of keys, right?

```suggestion
      my_params[:range].select do |solr_field, range_options|
        next unless range_options && blacklight_config.facet_fields[solr_field].present?
```

And, fwiw, this should be the behavior once we get range limit using components too.

dkinzer: Yeah, that makes total sense.  Thanks.

dkinzer:
This doesn't look like it no longer applies to the main branch.  I'm closing and reopening to apply to the 7.x-release branch.

dkinzer:
I'm putting this back into draft until I get a chance to see if the latest version fixes the issue I was having.

dkinzer:
The latest version of blacklight_range_limit does not have this issue.

PR #170, opened 2021-05-28
Use a blank facet.field solr param instead of deleting.

dkinzer:
When we outright delete solr_params["facet.field"] this causes Solr to
potentially use default "facet.field" configured in solrconfig.xml thus
nullifying trying to do a more efficient search.

By overriding solr_params["facet.field"] with a blank value then we
effectively override any default value set in solrconfig.xml


dkinzer:
Note: I didn't add a test for this, because that function has no tests associated with it.

jcoyne:
Is it possible that people depend on the existing behavior as a feature?

PR #179, opened 2021-12-11
Convert range limit views + helpers to presenters + components

cbeer:
Pending a release with https://github.com/projectblacklight/blacklight/pull/2594

This PR is perhaps not strictly backwards compatible for applications with extensive customizations, but is functionally equivalent and the public field configuration remains the same. Additionally, I've preserved the helper methods that were actively used by the old partials, so this should also be non-disruptive for applications that customized the partials. 

With that, and in the interest of keeping the major version number of this project similar to the blacklight major version number, I think there's a reasonable argument for just bumping the minor version in a release.

Upgrade guide
----

If you have no local customizations for the range limit plugin, there's no additional steps required.

The most common local customization seems to be overrides to combine multiple blacklight plugins that override the same upstream methods. 

This new approach should not require those overrides and the components and presenters should work for you as-provided. Make sure to remove any local `blacklight_range_limit/_range_segments.html.erb` view partial and consider updating your facet field configuration to the new recommended approach:

```ruby
config.add_facet_field 'pub_date', label: 'Publication Year', **default_range_config
```

Finally, if you have additional local customizations, these should be adapted into local presenters + components that extend and override the provided classes as appropriate. In general:

- overrides to the display of the selected range in constraints (likely by overriding the `range_display` helper) can be done with a custom FacetItemPresenter  
- overrides that control how range limit URL parameters are constructed can be done in a custom FilterField
- overrides that change how solr response information is mapped into blacklight can be done with a custom FacetFieldPresenter

Note: Some CSS classes and URL parameters have changed to match those provided by blacklight for ordinary facet fields.


lib/blacklight_range_limit/range_limit_builder.rb
```diff
@@ -12,59 +12,35 @@ module RangeLimitBuilder
...
-          solr_params[:fq] ||= []
-          solr_params[:fq] << "-#{config.field}:[* TO *]"
+      end
 
-        elsif !(hash["begin"].blank? && hash["end"].blank?)
-          # specified in request params, begin and/or end, might just have one
-          start = hash["begin"].blank? ? "*" : hash["begin"]
-          finish = hash["end"].blank? ? "*" : hash["end"]
+      ranged_facet_configs.each do |field_key, config|
+        solr_params["stats.field"] << config.field
```
barmintor, line 23: Since it's checked unconditionally later in the block, might clean things up to  `next if range_config[:segments] == false` at the top. Which also raises a question about whether to drop the stats assignment if there's no segmentable range facets.

spec/features/blacklight_range_limit_spec.rb
```diff
@@ -14,9 +14,8 @@
   it "should provide distribution information" do
     visit search_catalog_path
     click_link 'View distribution'
-
-    expect(page).to have_content("1500 to 1599 0")
-    expect(page).to have_content("2000 to 2008 12")
+    expect(page).to have_content("1500 to 15990")
```
barmintor, line 17:
```suggestion
    expect(page).to have_selector('a.facet-select', text: "1500 to 1599")
    c16 = page.find('a.facet-select', text: "1500 to 1599")
    expect(c16.ancestor('li')).to have_selector('span.facet-count', text: "0")
```

spec/features/blacklight_range_limit_spec.rb
```diff
@@ -14,9 +14,8 @@
...
+    expect(page).to have_content("2000 to 200812")
```
barmintor, line 18:
```suggestion
    expect(c16.ancestor('li')).to have_selector('span.facet-count', text: "0")
    c21 = page.find('a.facet-select', text: "2000 to 2008")
    expect(c21.ancestor('li')).to have_selector('span.facet-count', text: "12")
```

barmintor (approved):
Some minor performance and testing possibilities, but nothing major. I think there's something to be gained from minting a major rev here and moving on to bugfixes - this is generally the approach I also used. I would also like to open a related PR to the Blacklight 7.x branch to rely more on filters than direct inspection of `params[:f]` so that some of the deprecated methods would still work against this, but we can discuss separately.

barmintor:
Hmm failing on `net/smtp` w/ Ruby 3

jrochkind:
net/smtp moves from being a "default gem" to a "bundled gem" in ruby 3.0.  

https://stdgems.org/

"The behavior of bundled gems is similar to normal gems, but they get automatically installed when you install Ruby. They can be uninstalled and they are maintained outside of Ruby core."

What you are supposed to do about this in software that uses it as a dependency and is intended to work on multiple ruby versions.... continues to really confuse me.   I guess add it to your gem dependencies? Whether this will mess up older versions or ruby I don't know though? 

There is a long Issue on this... somewhere... not I can't find it, that at one point I tried to participate in, and I ended up just more really confused. Not ruby's strongest hour.

jrochkind:
Here's some of the long and confusing trail of people trying to deal with this. I think `mail` is in fact the dependency that's using net/smtp, relevant to us? https://github.com/mikel/mail/pull/1439

Which led to this...

https://github.com/mikel/mail/pull/1472

Looks like mail may finally be _about_ to solve the problem, and has a 2.8.0.rc1 release that does? But as an rc release, it won't be used here.  It may be the problem gets solved whenever `mail` 2.8.0 gets released. 

If ruby 3.0 inclusion in CI was not added in this PR but was pre-existing, I'm not sure why the problem is showing up only now. Very confusing.

jrochkind:
oh sweet, thanks for the info @barmintor. The matrix is locked to exact Rails patch version I guess? OK!

PR #195, opened 2022-07-13
Make range range limits scalar values

cbeer:


barmintor:
I am confident I was just trying to spec param parsing consistently across core blacklight and the plugins. I have no objection to changing this.

barmintor:
I kind of remember trying to figure out whether there was supposed to be a way to indicate more than one range?

cbeer:
No, I don't know of a way for a user to select multiple ranges for a single field.

barmintor:
I think this looks fine, but I'm spooked about approving this PR, community feedback suggests I don't know it as well as I think I do.

Here's what I will note:
- The current implementation supported arrays of values for the begin and end keys. This one does not. This might not be a use case. It's not in the UI.
- However we're supposed to know what Facebook might do with URLs, there's nothing here. I honestly don't know, I don't use Facebook and have no intention of ever logging into it to see how these things work.

I'm adding @jrochkind, who uses this plugin, as a reviewer. I don't think this change should be merged until he approves it.

barmintor (commented):
See above.

jrochkind:
I can confirm that this branch resolves the problems in my app, my app passes build again and seems to work properly/identically again with this branch, after previously breaking in an upgrade from blacklight_range_limit 8.1.0 to 8.2.2.

I can also confirm that blacklight_range_limit does not support selecting multiple ranges for a single field. 

I'm afraid I really don't understand what's going on in the code, I just don't have my mind wrapped around what's really going on here, what the code changes between 8.1.0 and 8.2.2 are doing, or what the code in this PR is doing. 

So I'm reluctant to approve it as a reviewer, I don't feel I have the competency to do that, I don't really understand whether or not this change might cause problems for people in the same way the original changes did, if this is reverting some aspect of the original changes? (I'm not sure it is!) Then could _this_ change cause problems for people for whom current 8.2.2 is working, to change it again? Maybe not? I just don't have a grasp of what's actually happening, at all, I'm sorry. 

If nobody else is available for review and you need me to approve it, maybe I could with more explanation of what's going on to be able to do so -- what led to the change I saw between 8.1.0 and 8.2.2, and how the code in this PR works. But I know that's hard work too. 

But I can confirm it works for my app! 

I appreciate the work both of you have done to help on this, @cbeer and @barmintor .

jrochkind:
OK, experimenting more, I can see that prior to this change, `blacklight_range_limit` had moved it's query params to always use an array format for range begin/end, eg (escaped) `"&range[year_facet_isim][begin][]=1900&range[year_facet_isim][end][]=1910"`

I am not totally sure if this was an intentional or not -- I would consider it unnecessary and undesirable, so maybe we could say it was a bug, and this PR fixes the bug?  If it's a bug fix, we don't need to worry if there are apps for whom *changing it back* would be disruptive, just like changing it was disruptive for me? 

I am not totally sure the right thing to do, and I myself don't understand the code in this PR.  But I will mark my "approval" on this PR indicating my support of whatever you two decide, @cbeer and @barmintor . 

I can make my app work again either way, with a new release with this PR -- or, if needed because it's decided the change in this PR is not desirable/acceptable, I can do the work to change my app to use the new form of query param with the extra array instead.  Just let me know, thanks!

jrochkind (approved):

jrochkind:
OK, I believe I've found another bug introduced by the "Array values" behavior previous to this PR. 

Let's say I do a range limit search, and my query params look like (escaped version of):

    range[year_facet_isim][begin][]=1900&range[year_facet_isim][end][]=1910

Now i'm looking at search results, with a range facet such as this:

Screen Shot 2022-07-19 at 2 26 30 PM

Let's say I want to modify my range limit, after looking at these results, I want to change it to `1905-1915` after all. If I make those edits in the box above, I expect them to _replace_ my previous `1900-1910` values. That is what they did in older version of blacklight/blacklight_range_limit. But with the "Array" values, if I enter new values in the boxes above and press "Apply", my new values get added on to the query params, with previous ones still there too:

     range[year_facet_isim][begin][]=1900&range[year_facet_isim][end][]=1910&q=jrochkind&range[year_facet_isim][begin]=1905&range[year_facet_isim][end]=1915&commit=Apply

If I keep editing, the query string will keep growing. 

While blacklight_range_limit is somehow doing the right thing and only paying attention to the "last" pair of begin/end in the query param, this seems unintentional behavior and kind of fragile. 

So I think that's another argument for merging this PR as a bugfix; alternately, if the "array" style params are to remain, there's a bug to be fixed: keep the query string from growing every time the user edits the range limits; it ought not to be keeping the previous inputs in the query string on every edit.

jrochkind:
I'm confused about how this is turning into something else, but I guess that's becuase I don't understand the change! 

@barmintor , @cbeer identified the change as coming from your #186.  What do  you think about that? If so, you would be the one who could answer if this change was done intentionally or not!  Either way, I _think_ you and cbeer are basically the only ones touching this stuff, so if neither of you made this change intentionally, then it probably was not done intentionally?

barmintor:
@jrochkind I got my notifications crossed, I thought we were getting feedback that this was an issue on 7.0.1

@jrochkind you said earlier that:

> OK, experimenting more, I can see that prior to this change, blacklight_range_limit had moved it's query params to always use an array format for range begin/end, eg (escaped) "&range[year_facet_isim][begin][]=1900&range[year_facet_isim][end][]=1910" -- I wasn't sure if it was just an issue of how params got serialized as hidden fields, but, nope, it was a change to always use/prefer query params in that format, with the final value being an array.

You seem to be saying that it was prior to #186 Can you identify the version where you saw it changed?

jrochkind:
Oh, okay crossed notifications confusing an already confusing situation! 

I did not mean to be saying that the change to make query param be an `[]` array happened prior to #186.  I noticed it happening somewhere between blacklight_range_limit  8.1.0 and 8.2.2 -- so that is consistent with it happening in #186, which seems to have dropped in 8.2.0. 

@barmintor , if you didn't introduce the change intentionally in #186, I think it's safe to say it was not introduced intentionally. 

So are we back to being good? 

I am not really sure how to write a spec... I guess it would be a spec for the nature of the params?  Anyone else have an idea?

barmintor:
@jrochkind you are observing that the plugin writes out URLs in a way that changed. If that is a bug, it is important that we nail that down in a spec. Your bug report is not about arrays or non-arrays working as pre-existing links, right? Just that the plugin is writing them out in the array style and ought not to?

jrochkind:
Yes, my bug report is about the query param names (in actual URL query string as well as in hidden input fields, where appropriate) changing from (in 8.1.0 and previous), eg:

    &range[year_facet_isim][begin]=1900&range[year_facet_isim][end]=1910

To (in 8.2.2)

    &range[year_facet_isim][begin][]=1900&range[year_facet_isim][end][]=1910

This change does not improve anything I am aware of, and it appears was not intentional?  In addition, along with this change is an additional bug I reported at https://github.com/projectblacklight/blacklight_range_limit/pull/195#issuecomment-1189423243, where editing the begin/end on an existing constraint results in the values showing up in URL query param twice, with different forms, and us counting on the "last" one being used which also seems to be a fragile accident of implementation. 

So my hypothesis is that this change is a bug which should be reverted. 

I am not certain how to write a test for it. Does anyone else have an idea?

barmintor:
@jrochkind I am trying to identify the gap and write the spec.

barmintor:
Ok, so I think it's the case that the form in the sidebar, which uses a locally defined helper, has *not* (in the interim) writing the final brackets (and this is what is tested) - but the search param serialization, used in "Back to Search" links and when adding a value to a *different* facet, was writing the array-indicating brackets (untested). So there's the gap in specification.

jrochkind:
Thanks @barmintor . I feel like that further supports the idea that the change was a bug, being reverted here. The two forms should match. 

I won't have any time this week, but next week I can see if I can figure out a way to test this too, if you haven't. 

However, I think even if nobody has time to figure out the testing, we should still revert the bug/change, and the sooner the better hopefully reducing the chances anyone comes to rely on the buggy behavior.

jrochkind:
Hey @barmintor, I've spent a couple more hours on this, but have been unable to come up with a spec. 

I keep thinking I'm close, but I can't manage to get a spec that produces the wrong `[]` in param names (eg in hidden inputs) in main. I can't actually reproduce this -- my specs still don't have the `[]` on there. So i can't create a spec that fails on main but passes only with this fix from cbeer. 

I feel like I've gone down some pretty deep holes of abstraction, jumping between code defined here and in blacklight proper, and just gotten lost. The only change in this PR seems to be in the `BlacklightRangeLimit::FilterField#values` method...  but that method **returns** a range value like `[2013..2022]`, and returns the same thing in master and this branch for the input I've figured out to throw at it... I can't understand how the change in that method even results in the changed behavior we are seeing. 

And yet, the branch does make a difference in my app, avoiding the spurious/unintended/buggy `[]` params. 

So... I'm a loss. I don't think I'm going to have enough time to get to the bottom of this. Unless @cbeer wants to give us a test case... @cbeer seems to be okay with this PR without a test, I am going to be reluctantly okay with it because I want the fix and don't see another option, what do you think, can we merge the fix without a spec change?

barmintor:
Having strong expectations about the URL as versioned API and patching without specs seem to be at odds to me?

jrochkind:
I'm sorry, I'm not following. Or maybe I agree it is "at odds" if that means "undesirable", but I don't know what to do with it?  I have been unable to come up with a test. 

I made my case that this is a bug, I think, but I can do it again if it's useful. 

What is your opinion on whether the current behavior is a bug, and on what we should do about it either way? Like, you could suggest we close this PR from cbeer, and just leave blacklight_range_limit as it is?  

But... it definitely seems to be a bug that you get double params when you edit existing limits, right?  This PR fixes the bug. @cbeer who I am grateful to for providing this PR didn't provide a spec with it. Neither you nor I have been able to come up with one either. 

So... now what?  I would propose fixing the bug without a spec, becuase we lack the resources to come up with one, but should not let that keep us from fixing the bug. Are you blocking? Or making an alternate proposal?

barmintor:
ok, forgive the minor history rewrite. I'm bumping the minimum Blacklight version to 7.25.2 so that we can the necessary param sanitization behavior we want to test - otherwise I think this catches it.

barmintor (approved):

jrochkind:
Thank you @barmintor!

jrochkind:
OK, testing this branch (or now master) again in my app...  I'm afraid it's still adding the double params when you edit the range limit in the panel, something it was not doing in blacklight_range_limit 24.  :(.   This is the bug i was talking about in https://github.com/projectblacklight/blacklight_range_limit/pull/195#issuecomment-1189423243, which I think is a bug. 

I will file an issue on the bug I am talking about. I am off tomorrow, so may not have a chance until next week, but I'll try to get it filed today before I leave.

jrochkind:
Oh my, spending yet more time with it... the behavior I am seeing as a bug reproduces pre-Blacklight 7.25, it's actually not related to this stuff. Sorry, the different versions of blacklight and blacklight_range_limit involved get confusing and hard to keep track of and set up reproductions for. I will still file this as an issue.

PR #198, opened 2022-07-28
do not render a range form's field value as hidden and visible inputs

### LINKED ISSUE #197: Double params get included in query string when editing range limits
This reproduces off recent main branch at 204cb54c3d00, blacklight 7.29.0

1. I apply a range limit, 1960-2006
    * The query string looks like this (unescaped): `range[pub_date_ssim][begin]=1960&range[pub_date_ssim][end]=2006&commit=Apply`
     *  Screen Shot 2022-07-28 at 4 25 33 PM

2. I decide to modify my limits, by changing the numbers in the input panels in the range limit sidebar, and clicking apply. I change to 1970 and 2000. 
   * There are now *two sets* of params in my query param: `range[pub_date_ssim][begin]=1960&range[pub_date_ssim][end]=2006&range[pub_date_ssim][begin]=1970&range[pub_date_ssim][end]=2000&commit=Apply`. **this is what I am suggesting is a bug**
   * The "last" one, that I choose most recently, is the one that gets "picked up" by the app, so that's good at least....
   * Screen Shot 2022-07-28 at 4 29 00 PM

3. What happens if I edit again? This time to 1975 and 1995
    * Still only two sets of params in the query string, it doesn't grow forever, but just keeps the previous one in there
    * `?range[pub_date_ssim][begin]=1970&range[pub_date_ssim][end]=2000&range[pub_date_ssim][begin]=1975&range[pub_date_ssim][end]=1995&commit=Apply`

# When did this bug appear?

At first I thought this was new with the changes in Blacklight 7.25. 

But it turns out this same buggy behavior does exhibit (I think) in blacklight `7.24.0` and blacklight_range_limit `8.1.0`. I don't know how much earlier before that. 

There is an intermediate period, eg blacklight 7.29.0 and blacklight_range_limit 8.2.2 where the problematic behavior still reproduces, but even more confusingly with two different versions of query params, one with a trailing `[]` (`range[pub_date_ssim][begin][]`) and one without (`range[pub_date_ssim][begin]`), as in: `range[pub_date_ssim][begin][]=1960&range[pub_date_ssim][end][]=2006&range[pub_date_ssim][begin]=1970&range[pub_date_ssim][end]=2000&commit=Apply"` #195 restores the previous behavior, where the double params both have the same key, `range[pub_date_ssim][begin]`

barmintor:
- fixes #197

### DIFF
--- a/app/components/blacklight_range_limit/range_form_component.html.erb
+++ b/app/components/blacklight_range_limit/range_form_component.html.erb
@@ -5,8 +5,8 @@
     <%= render_range_input(:begin, begin_label) %>
     <%= render_range_input(:end, end_label) %>
     <div class="input-group-append visually-hidden">
-      <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit] %>
+      <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit], name: nil %>
     </div>
-    <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit] + " sr-only", "aria-hidden": "true" %>
+    <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit] + " sr-only", "aria-hidden": "true", name: nil %>
   </div>
 <% end %>
--- a/app/components/blacklight_range_limit/range_form_component.html.erb
+++ b/app/components/blacklight_range_limit/range_form_component.html.erb
@@ -1,5 +1,5 @@
 <%= form_tag search_action_path, method: :get, class: [@classes[:form], "range_#{@facet_field.key} d-flex justify-content-center"].join(' ') do %>
-  <%= render Blacklight::HiddenSearchStateComponent.new(params: @facet_field.search_state.params_for_search.except(:utf8, :page)) %>
+  <%= render Blacklight::HiddenSearchStateComponent.new(params: @facet_field.search_state.params_for_search.except(:utf8, :page).tap { |p| p[:range]&.except!(@facet_field.key) }) %>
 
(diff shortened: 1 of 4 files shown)


app/components/blacklight_range_limit/range_form_component.html.erb
```diff
@@ -1,12 +1,12 @@
 <%= form_tag search_action_path, method: :get, class: [@classes[:form], "range_#{@facet_field.key} d-flex justify-content-center"].join(' ') do %>
-  <%= render Blacklight::HiddenSearchStateComponent.new(params: @facet_field.search_state.params_for_search.except(:utf8, :page)) %>
+  <%= render Blacklight::HiddenSearchStateComponent.new(params: @facet_field.search_state.params_for_search.except(:utf8, :page).tap { |p| p[:range]&.except!(@facet_field.key) }) %>
```
cbeer, line 2: I wonder if we should pull the params (or the whole `HiddenSearchStateComponent.new`) up to the component class? At least that way, we could leave a good comment on the method about what's going on.

barmintor: that would make sense to me - I'm maybe inclined to just haul the component up, for the reasons you bring up and also because `<%= render hidden_search_state %>` seems nice for the template.

barmintor: though just doing the params would be a nicer test. Hmm. What do you think?

cbeer: `<%= render hidden_search_state %> ` looks pretty good to me.

jrochkind:
Thank you @barmintor and @cbeer , much appreciated.

PR #200, opened 2022-08-06
Do not render constraint for un-configured field

### LINKED ISSUE #168: Range limit constraint does not take into account blacklight config

dkinzer:
Fixes #168

Fixes issue where a constraint gets rendered even though the controller
should not know anything about the particular field whose constraint is
being rendered.

(This is a redo of #169 for the release-7.x branch).


dkinzer:
Actually I need to rethink this.

Issue #187, opened 2022-05-19
Updating begin value increments end value


barmintor:
Reported by Tao Zhao on the Code4Lib Slack:

> A bug with black_range_limit 7.0 plugin:
After I click the button "Limit" to filter the results, and then if I move the start year along the slider, the end year will automatically add by one, i.e., if I search by 2000-2022, then change the 2000 on the left side to be 2005, the year 2022 on the right side will change into 2023, automatically.
Before I do the search using 2000-2022:
![image](https://user-images.githubusercontent.com/119860/169411933-5fb18e08-07b6-4f67-aa69-d70ee40d3892.png)

> Then I change the start year to be 2005, the end year automatically adds by one to be 2023:

![image (1)](https://user-images.githubusercontent.com/119860/169412020-230fc70f-33c8-4483-a37f-2c7250aa2fa7.png)

> I've found some other institutions also had this issue and I wonder if this has been fixed in some later versions. Thanks.

barmintor:
it looks like the change over to the bootstrap slider way back in version 5 (2014) has an alteration of the begin element change handler: https://github.com/projectblacklight/blacklight_range_limit/commit/fa965a73594ac58894add7056235c0895f296a28

if I'm reading that right, it formerly did a insertion-style update of the begin value, but afterwards did an update of the entire value array, the second element of which has been offset by 1 when the slider is constructed, presumably to be inclusive?

barmintor:
Regardless, I am struggling to reproduce the described behavior on the demo app against Blacklight 7.25.2.

jrochkind:
I *did* reproduce on my own app.  Blacklight 7.24.0, blacklight_range_limit 8.1.0. (Have not debugged, just reproduced). 

I can't tag Tao the reporter because this was proxy-reported by barmintor, but Tao, if you see this, it would be helpful to know what version of Blacklight and blacklight_range_limit your report is from.

MPLSFedResearchTZ:
Could you please make corresponding changes in the version 7.0.1 too? The reason is we are working on the Hyrax 2.9.0, which uses blacklight 6. We would really appreciate that if you make patches to Blacklight_range_limit 7.0.1 too to fix the end limit value issue.

PR #201, opened 2022-08-08
Stop trying to tweak the end value

MPLSFedResearchTZ:
This is to apply the fixes of https://github.com/projectblacklight/blacklight_range_limit/issues/187 to v7.0.0.


app/assets/javascripts/blacklight_range_limit/range_limit_distro_facets.js
```diff
@@ -245,11 +245,11 @@ Blacklight.onLoad(function() {
           var values = $(event.target).data("slider").getValue();
           form.find("input.range_begin").val(values[0]);
           form.find("input.range_end").val(values[1]);
-          plot.setSelection( normalized_selection(values[0], Math.max(values[0], values[1]-1)), true);
+          plot.setSelection( normalized_selection(values[0], Math.max(values[0], values[1])), true);
```
cjcolvar, line 248: The upstream commit calls `normalized_selection` explicitly with `BlacklightRangeLimit.normalized_selection`.  Do you think we should do that here and on line 252 below?

cjcolvar (commented):
This looks like a good start, but I'm only seeing the changes from the first commit from #191 backported here.

It looks like you might be manually copying the changes.  I find it easier to use `git cherry-pick` and then fix any conflicts.  This helps preserve the semantics of the original commits as well as the authorship.  It is also a time saver when backporting multiple commits.

app/assets/javascripts/blacklight_range_limit/range_limit_distro_facets.js
MPLSFedResearch: In v7.0.0, normalized_selection is a function defined in file range_limit_distro_facets.js and calling it in range_limit_distro_facets.js does not require the "BlacklightRangeLimit" namespace. This "BlacklightRangeLimit.normalized_selection" is only defined in version 8 in the file of range_limit_shared.js.

cjcolvar (dismissed):
This looks good to me.  I switched the PR base to be a new `release-7.0.x` branch since this is targeting a 7.0.1 release.  Switching the base resolved the conflicts.

cjcolvar (approved):

PR #208, opened 2022-10-03
Don't load Blacklight config until the generators have run

jcoyne:
Otherwise you will encounter an error about ::SearchBuilder not having been defined yet


cbeer:
Alternately, have Blacklight::Configuration not blow up? https://github.com/projectblacklight/blacklight/pull/2822

jcoyne:
Preferred BL#2822

PR #223, opened 2022-11-15
CSS file is not .scss and does not use any .scss features, should not end in .scss

jrochkind:
This file only uses a sprockets 'require'. It does not use any scss features, and never has, since it's introduction in 8397f56112

I think it was a mistake initially to name it .css.scss. As we start exploring differnet ways to manage assets in the brave new world, it ending in .scss makes things more confusing, implying things about it that are not true -- that it needs to be @imported in an SCSS pipeline, or that it CAN be imported in an .scss pipeline (it won't work, as it the sprockets require won't be triggered if you do it that way!)

This should not disrupt anything, as the require apps use is just `require  'blacklight_range_limit'` anyway, doesn't actually have the .scss on it (and actually wouldn't work if it did), rails has just been somehow covering up for our mistake all along, but will work fine with it fixed.


jrochkind:
Test failure in ruby 3.1 only, I believe is unrelated to this PR. I believe main branch would currently be failing in 3.1 as well. 

```
TypeError:
  no implicit conversion of nil into String
# ./.internal_test_app/config/environment.rb:5:in `<top (required)>'
# ./spec/spec_helper.rb:4:in `<top (required)>'
# ./spec/components/range_facet_component_spec.rb:3:in `require'
# ./spec/components/range_facet_component_spec.rb:3:in `<top (required)>'

An error occurred while loading ./spec/components/range_form_component_spec.rb.
Failure/Error: EngineCart.load_application!

FrozenError:
  can't modify frozen Array: ["/home/runner/work/blacklight_range_limit/blacklight_range_limit/.internal_test_app/app/controllers", 
```

See also #222, which is failing with same error, although it's oddly in ruby 3.0 rather than 3.1 (some race condition?)

jcoyne (approved):

PR #226, opened 2023-01-24
Publish assets to NPM

jcoyne:
This works best for javascript tools like sass, webpack, and esbuild

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "blacklight-range-limit",
   "version": "8.2.4-beta.1",
   "description": "A range facet UI component plugin for blacklight",
-  "main": "app/assets/javascripts/blacklight",
+  "main": "app/assets/javascripts/blacklight_range_limit",
   "scripts": {
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,0 +1,29 @@
+{
+  "name": "blacklight-range-limit",
+  "version": "8.2.4-beta.1",
+  "description": "A range facet UI component plugin for blacklight",
+  "main": "app/assets/javascripts/blacklight",
+  "scripts": {
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/projectblacklight/blacklight_range_limit.git"
+  },
+  "files": [
+    "app/assets", "vendor/assets"
+  ],
+  "author": "",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/projectblacklight/blacklight_range_limit/issues"
+  },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "blacklight-range-limit",
+  "version": "8.2.4-beta.1",
+  "description": "A range facet UI component plugin for blacklight",
+  "main": "app/assets/javascripts/blacklight",
```
barmintor, line 5: Should this be blacklight_range_limit?
```suggestion
  "main": "app/assets/javascripts/blacklight_range_limit",
```

package.json
```diff
@@ -0,0 +1,29 @@
...
+    "url": "https://github.com/projectblacklight/blacklight_range_limit/issues"
+  },
+  "homepage": "https://github.com/projectblacklight/blacklight_range_limit#readme",
+  "devDependencies": {
+  },
+  "browserslist": [
+    "defaults",
+    "not IE 11"
+  ],
+  "dependencies": {
```
barmintor, line 27: Do we need to document the flot and bootstrap-slider dependencies here?

barmintor (commented):
I don't know if there's some support for doing a sprockets build as part of the npm build & push, so these are more questions than comments.

package.json
jcoyne: These are in the vendor/directory, so probably not?

barmintor: Oh, I see. But jquery isn't, which would bomb in a default BL8 app, right?

barmintor: Right but installing from npm should also work if we do this. I'm going to try generating the test app with something besides sprockets to see what happens.

barmintor (approved):
Well, now I know what happens when you don't use sprockets, but that's out of scope here

package.json
jrochkind: recommend listing jquery as a peerDependency if you do list it. If you list it as a straight dependency, npm will _insist_ on installing it -- even if the installation preferred to provide it with sprockets, as jcoyne I think was referring to. 

Also jquery as a dependency it's very easy to end up with npm having multiple versions of jquery installed at once, which is kind of annoying and hard to figure out what version of jquery gets used where (and I think some risk of including jquery multiple times in your bundle!)

Even listing as a peer dependency, it's going to be annoying for people who want to install jquery via sprockets instead. 

In general I'm totally not sure what the right thing to do is -- I've just experienced the annoyances that various things (maybe all the things!) can cause!  I guess I'd say be careful making any changes unless they are to address a specific known current issue, and ideally unless you test under various JS packaging scenarios (which is a big pain; I think at some point BL is going to have to be more limited about what JS packaging scenarios it supports).

PR #228, opened 2023-01-31
Commenting out code that is making this plugin fail in blacklight 8

carolyncole:
This is not the final solution, but I am putting it on github so others may have access to it...


jrochkind:
No longer relevant to current code, 9.0 works with BL 8!

PR #239, opened 2023-08-11
use modern ViewComponent slot syntax as Blacklight

### LINKED ISSUE #238: Facet field partial does not use modern slot syntax
The component doesn't render the `body` or `label` slots, and thus the component specs don't pass.

barmintor:
- remove stub of removed method from presenter mock in spec
- fixes #238

### DIFF
--- a/app/components/blacklight_range_limit/range_facet_component.html.erb
+++ b/app/components/blacklight_range_limit/range_facet_component.html.erb
@@ -1,9 +1,9 @@
 <%= render(@layout.new(facet_field: @facet_field)) do |component| %>
-  <% component.label do %>
+  <% component.with_label do %>
     <%= @facet_field.label %>
   <% end %>
 
-  <% component.body do %>
+  <% component.with_body do %>
     <div class="limit_content range_limit <%= @facet_field.key %>-config blrl-plot-config">
       <% if @facet_field.selected_range_facet_item %>
         <%= render BlacklightRangeLimit::RangeSegmentsComponent.new(facet_field: @facet_field, facet_items: [@facet_field.selected_range_facet_item], classes: ['current']) %>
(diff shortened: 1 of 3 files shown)


jrochkind:
Thanks for this @barmintor .  Before seeing you had done this, I went ahead with #240, more or less duplicating this. 

This is necessary to use blacklight_range_limit with BL7 and view_component 3, which I would really like to do. 

I think these failures are for unrelated reasons. I don't know how to fix them either, it seems tricky. 

What if fixing the unrelated build errors is not a pre-requisite for fixing this PR? 

These are small and targetted changes. I have tested the changes locally, blacklight_range_limit definitely still works (as long as you have view_component >= 2.54.0). 

Thoughts on how to proceed?

jrochkind (approved):

jrochkind:
Somehow the .internal_test_app Gemfile is being generated with a corrupt gem version specification. This may be an engine_cart issue? I haven't been able to figure out where it's coming from or how to fix it.

jrochkind:
i can't get tests to run locally for other reasons, that are not this failure. When I try to run tests locally with `bundle exec rake ci`, I get the old:



```
Failure/Error: EngineCart.load_application!

FrozenError:
  can't modify frozen Array:  [...]
# /Users/jrochkind/.gem/ruby/3.2.2/gems/railties-7.0.8/lib/rails/engine.rb:574:in `unshift'
```

Which I have run into before, and don't know what to do with it.

jrochkind:
OK, I spent today figuring out why the build here was broken in this way, and filed this PR on blacklight:

https://github.com/projectblacklight/blacklight/pull/3088

However, even after fixing that, the build STILL does not pass with Blacklight 8.  It fails with other errors. I don't believe the build has ever passed on BL8. BUT... there are people who are using blacklight_range_limit on BL8 already anyway... 

I will file another PR to no longer test on BL8, so the build can be green again?

jrochkind:
dup, handled in #242 instead

PR #243, opened 2023-10-12
add direct view_component dependency

jrochkind:
With #242 to use Slots v2 API, we now require at least view_component 2.54.0, that's when it was introduced. https://viewcomponent.org/CHANGELOG.html#2540

I think we learned from #242 (and analagous things happening with other BL plugins) that if we're using ViewComponent API directly, as we are, we should declare a direct dependency on it, and what we know to work with, instead of just relying on indirect dependency from blacklight. So also adding a cap at `< 4`, since a future 4 might include currently unknown backwards incompat changes that effect this code. I think this is just responsible dependency declaration.

@jcoyne?


jcoyne (approved):

jrochkind:
Thank you @jcoyne !

PR #245, opened 2023-10-13
Webdrivers is no longer needed

jcoyne:


jrochkind:
If CI passes without it, good enough for me!

PR #246, opened 2023-11-02
bump version to 8.4.0 for release

jrochkind:


jrochkind:
I understand community norms are I can self-merge a version-bump-only PR with green CI -- someone tell me otherwise and I won't do it again, but for now doing it!

PR #248, opened 2023-12-01
fix: add mutation observer for BL8 Modal and support Turbo

yetti:
Blacklight 8 uses an HTML Dialog element instead of a Bootstrap modal, so it won't fire the 'shown.bs.modal' events when opened. This adds a MutationObserver to check if the 'open' attribute is added to the HTML Dialog element, to determine if it's open. It also adds a function that can be called to add 'turbo:load' event handlers for those that want to support re-/rendering after Turbo navigation.


yetti:
Closing as there are more issues around supporting Turbo than I initially thought. It's probably better not to include that functionality in this PR.

PR #249, opened 2023-12-01
add MutationObserver for Blacklight 8 Modal

yetti:
Blacklight 8 uses an HTML Dialog element instead of a Bootstrap modal, so it won't fire the 'shown.bs.modal' events when opened. This adds a MutationObserver to check if the 'open' attribute is added to the HTML Dialog element, to determine if it's open.


jcoyne:
Would it be simpler to have Blacklight itself emit an event that we can bind to here?

jcoyne (approved):

PR #257, opened 2024-06-20
Remove aria-hidden from visible submit button

taylor-steve:
Causes:

Screenshot 2024-06-20 at 11 34 48 AM

### DIFF
--- a/app/components/blacklight_range_limit/range_form_component.html.erb
+++ b/app/components/blacklight_range_limit/range_form_component.html.erb
@@ -7,6 +7,6 @@
     <div class="input-group-append visually-hidden">
       <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit], name: nil %>
     </div>
-    <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit] + " sr-only", "aria-hidden": "true", name: nil %>
+    <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit] + " sr-only", name: nil %>
   </div>
 <% end %>


app/components/blacklight_range_limit/range_form_component.html.erb
```diff
@@ -7,6 +7,6 @@
     <div class="input-group-append visually-hidden">
       <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit], name: nil %>
     </div>
-    <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit] + " sr-only", "aria-hidden": "true", name: nil %>
+    <%= submit_tag t('blacklight.range_limit.submit_limit'), class: @classes[:submit] + " sr-only", name: nil %>
```
jcoyne, line 10: `sr-only` was the class to use to hide things in bootstrap 4.  In Bootstrap 5 that was renamed to `visually-hidden`.  In order to keep compatibility with both, we should have both of those classes here.

jcoyne (changes requested):

app/components/blacklight_range_limit/range_form_component.html.erb
taylor-steve: Ah I see. Perhaps they (`sr-only`, `aria-hidden`) shouldn't be on that element, but the `visually-hidden` element above?

taylor-steve: Although in that case I don't know why the first submit has `input-group-append`. Visually it seems like the second submit button is intended to be the visible button:

Screenshot 2024-06-20 at 3 29 03 PM

If I swap all the hidden classes to the second submit tag it looks like:
Screenshot 2024-06-20 at 3 28 45 PM

Tests pass either way...

taylor-steve: Okay I keep forgetting that `input-group-append` doesn't exist in BS5. It seems like:

* the submit button within `input-group-append` is the intended submit button for BS4 (hidden in BS5 by `visually-hidden`)
* the second submit button is the intended submit button for BS5 (hidden in BS4 by `sr-only`). 
 
In which case, the original class configuration seems correct to me, it's just a matter of if we can safely drop the aria-hidden when on BS4.

jcoyne: @taylor-steve It doesn't look like you are getting the javascript activated correctly. Perhaps that has something to do with what you are seeing?

jcoyne: It looks like this was added in https://github.com/projectblacklight/blacklight_range_limit/commit/3f74bcedacd96731f7c26d3f4f97cf95f2ef5a1b#diff-f5ae40eafb6107f25255824abb5e99677cc38580c6cc3f9385bc574b3af8bb6aR67 and looks like the aria-hidden may have been added in mistake.

jcoyne: Okay, now I see what's going on.  There are two buttons. One was intended for BS4 and one is intended for BS5.  I'm not sure we can properly support accessibility and support both BS4 and BS5.

jcoyne: @taylor-steve The only thing I can think to do is to make a local version of this component that doesn't have a bs4 button on it.

PR #261, opened 2024-09-25
Pin to solr version 9.6.1

jcoyne:


jrochkind:
For the record, I believe this is because at this point latest `solr_wrapper` 4.0.2, which is used in CI, is incompatible with Solr 9.7+?

This pin can probably be removed if solr_wrapper is fixed. See currently red https://github.com/cbeer/solr_wrapper/pull/154

While this can make CI work again, this may still be very confusing for depending developers if they are trying to use solr_wrapper.

jrochkind:
This PR also contains code that looks like intended to support propshaft... is that intended @jcoyne or accidentally included?

jcoyne:
@jrochkind Yeah. That was a mistake.  I've updated so it's just the solr version change.

PR #262, opened 2024-09-25
Test on latest three versions of Ruby & Rails

jcoyne:


jrochkind:
Very nice thanks

jrochkind:
Oh wait, it was a green button so I merged, but did this actually run CI??? Now i"m not seeing the record so am not sure?

PR #263, opened 2024-09-25
Start solr with docker compose

jcoyne:


jrochkind:
I'm very novice at docker/docker-compose. 

Woudl it make sense to add something to README in the testing section. Currently it says (perhaps already out of date), 


> Test coverage is not great, but there are some tests, using rspec. Run bundle exec rake ci or just bundle exec rake to seed and start a demo jetty server, build a clean test app, and run tests.

(Is that still all that you need to do to run tests?)

With different paths in "local testing" sub-section too. 

https://github.com/projectblacklight/blacklight_range_limit?tab=readme-ov-file#tests

Can someone that understands better make sure that's all accurate and up to date and all anyone needs to know to run tests with this PR, before merging? Thank you!

jcoyne:
@jrochkind yes, that's all you need to do. This is the same approach used by blacklight core.

jrochkind:
Thanks! I will try testing it, since I am having trouble getting tests to run locally in master, but I may not be able to get it to work for similar unrelated reasons.

Also, perhaps a note you need Docker or equivalent (I am not sure what equivalents might be) installed and running?  That is a difference as of this PR?

jrochkind:
OK, I think maybe the instructions in README in [Local testing](21567) are already out of date and don't work maybe? Very confused about what's going on in testing world and how to get local testing to work!

https://github.com/projectblacklight/blacklight_range_limit?tab=readme-ov-file#local-testing


> If you want to iterate on a test locally and do not want to rebuild the required test environment every time you run the test you can set up the required server by first running:
>
>       bundle exec rake test:server
>
>Now from another shell run your individual test as needed:
>
>       bundle exec rspec spec/features/blacklight_range_limit_spec.rb
>
> Once you are done iterating on your test you will need to stop the application server with Ctrl-C.

Issue #264, opened 2024-09-25
Rewrite JS from the ground-up, for a major version "reboot"


jrochkind:
Current JS is over 10 years old, doesn't represent best practices, and has become increasingly hard to integrate with recent Blacklight and Rails. 

For people who don't want the graphical histogram, my understanding is that some are working on having just text entry "start" and "end" independent of this gem, based on features already added to blacklight itself -- which makes sense to me, just text entry can be much simpler than this gem. 

For those who want the histogram (my stakeholders do)... 

I would like to rewrite it from scratch, intending on a "reboot" of blacklight_range_limit with a new major version, and possibly some changes/loss of front-end UI functionality. 

* rewrite JS from scratch using low-dependency JS
  * Using [chart.js](https://www.chartjs.org/) (popular, maintained, plugin ecosystem, supports requires of just needed func for tree-shaking bundlers)
  * Delivered as an npm package, intended to be used with any JS bundling/delivery technique that can use npm dependencies, which should be everything in Rails-land but sprockets. (not focusing on importmaps, it should work _treating our npm package like any other_, not with Rails engine specific features, but may run into trouble if you have complicated dependency tree, because that's importmaps)
   * Definitely no JQuery
   * No npm dependencies but chart.js (which currently has one dep with no deps of it's own)

* Try to figure out right features to make sure the existing hidden textual description of facet bins is properly marked up for low-vision accessibilty -- might be as simple as leaving a textual list of bucket facets there as sr-only?

* With rewrite from scratch, some features may not make it into first version. 
   * Probably histogram and text entry boxes only
   * No range slider
   * No drag to select on histogram
   * No "view in larger window"

* I will attempt to have NO CSS provided by gem, as CSS dependencies from npm packages are still handled less consistently by various bundlers and are not supported by importmaps. (chart.js uses canvas and needs no CSS)

* If it will be a major version, defintely remove old ruby and Rails and Blacklight versions from support

* Obv need to end with CI working

jrochkind:
Closed by #266 and subsequent

PR #266, opened 2024-10-02
rebuild JS from ground up, prep major version

jrochkind:
For #264


app/assets/javascripts/blacklight-range-limit/index.js
```diff
@@ -0,0 +1,283 @@
...
+        });
+    } else {
+      handleOnPageData();
+    }
+  }
+
+  // Extract our bucket ranges from HTML DOM, and store in our instance variables
+  extractBucketData(facetListDom = this.container.querySelector(".facet-values")) {
+    this.rangeBuckets = Array.from(facetListDom.querySelectorAll("ul.facet-values li")).map( li => {
+      let from    = this.parseNum(li.querySelector("span.from")?.getAttribute("data-blrl-begin") || li.querySelector("span.single")?.getAttribute("data-blrl-single"));
```
jcoyne, line 109: can't these be const?

app/assets/javascripts/blacklight-range-limit/index.js
```diff
@@ -0,0 +1,283 @@
...
+    return undefined;
+  }
+
+  setupDomForChart() {
+    if(this.chartCanvasElement) {
+      // already there, we're good.
+      return this.chartCanvasElement;
+    }
+
+    let listDiv = this.container.querySelector(".facet-values");
```
jcoyne, line 152: this also seems like a const

jcoyne (commented):
Great! I'm so excited for this work!

jrochkind:
Thanks @jcoyne . Definitely still a work in progress. Can't figure out why I can't get Capybara tests to run on github CI (they run fine for me locally); have to go back to main branch and verify they really were running there I guess. 

I wans't thinking too much about let vs const, does that actually make a big difference? Happy to fix those though once I actually get tests working, if you have any ideas to share from your struggles with it please feel free.

jrochkind:
OK I'm getting closer. 

1. I tried to switch to `system` spec instead of `feature`, which I use in my own apps. But maybe it doens't work with engine_cart?  It was not succesfully starting headless chrome at all when doing it with system specs, fine turned back to feature, don't need to fight with that right now.   Don't understand why this wans't a problem running tests locally. 

2. I tried to switch away from a custom-defined selenium driver -- but I think maybe we need one to increase screen size, as at default driver screen size the facets collapse to be behind a menu, which is not what we write tests for?  Switched that back for now too.  Don't understand why this wasn't a problem for me running tests locally though. 

Still have one test failing that is NOT failing locally, and I don't know why. It could be a JS race condition of some kind... hard to debug when it's not happening locally (and tests take so long to run on github, with setup etc), may need to me to make several commits to keep testing things, may do that here, or in a throwaway branch.

jrochkind:
Tests are now *running*. 

But the test checking to make sure chart.js actually added a `canvas` to DOM is failing in CI while passing locally. So perhaps the JS isn't really working when run in CI?

I wonder if there's an easy way to get access to that screenshot on failure. 

All the self-deprecation messages in logs don't help, but want to get tests passing before I try to remove deprecated stuff. 

Definitely at least 50% of this task is messing with infrastructure/CI, rather than actual func.

jrochkind:
OK it's not a test/capybara issue -- our JS really isn't being installed and run correctly on CI (the test failure is real!) even though it is locally.

jrochkind:
omg. right. Because the blacklight generator, when detecting sprockets, isn't generating what I need it to into application.js... not sure why it is on my CI though with the same rails versions!  

We are not supporting pure sprockets, although we support sprockets+importmap no problem.

jrochkind:
OK I _think_ you can use Blacklight7 via importmaps -- you definitely can use blacklight 7 via jsbundling-rails. BUT the Blacklight 7 generator won't generate such an app. Which makes things challenging for CI based on the engine_cart generation approach. 

I really want this to work for BL7, which means it should run CI for BL7, will have to think more about how to best accomplish.

app/assets/javascripts/blacklight-range-limit/index.js
jrochkind: done!

app/assets/javascripts/blacklight-range-limit/index.js
jrochkind: done!

jrochkind:
This is ready for some alpha-testers. 

People who expressed _potential_ interest in alpha-testing include @dkinzer @seanaery @bess and @tpendragon . Anyone is welcome!

Right now this is unreleased so you have to install from git branch, currently under my account. `https://github.com/projectblacklight/blacklight_range_limit/`, branch `reboot_js_etc` -- if this is a barrier this could be released as a 9.0.0.alpha1, but I'm not totally sure if we have consensus on going forward with this as a 9.0.0 so was holding off (as well as on merging this). 

There are two main areas of testing:

1. Installation process -- Can you get it actually installed in your app? Easily?

2. UX/UI, including:
  * front-end Features removed from version 8, are there any that are highly desirable to bring back, or blockers for you?
  * Accessibility (to low vision) -- the categories in the histogram are currently being made available in a hidden screen-reader-only block.  But specific feedback on any aspect of accessibility or advice on what it needs to look like to be better for low-vision accessibility is very welcome. 

It would be great if anyone let me know if they are alpha testing; and also please feel free to get in touch (slack is great, or here) if run into any trouble, and with feedback!

bess:
Thank you! I will take a look asap

seanaery:
Really nice work, @jrochkind . I was able to try this branch out on our ArcLight app (BL8, importmaps, Bootstrap 5). It worked great. I have notes and suggestions and a screenshot over here: https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/merge_requests/380 

I should be able to do a similar trial run with a BL7, jsbundling-rails, Bootstrap 4 app later this week...

jrochkind:
Thanks @seanaery !

That missing apply button keeps coming and going for me, I have seen it before but have a hard time reproducing it. Thanks for reporting it, and that it was Bootstrap 5 that reproduced it for you. It definitely has _something_ to do with the code (already there, not added by me) trying to support both Bootstrap 4 and 5 in that inline input group. 

I don't totally follow your second issue, probably because i don't use plugins that have those conditions  -- can you do me a favor and see if it's pre-existing with the current blacklight_range_limit?  My *guess* is that this is a pre-existing condition before and after this PR.  Since this PR pretty much _only_ touches front-end display such as histogram, and nothing on the back-end that calculates buckets and numbers. 

If it's pre-existing, then I think it's out of scope for this PR to fix it, but.. if you wanted to PR a fix, I'd spend time reviewing/merging!

If it was caused by this PR and didn't exist previous to it, then I would want to fix it before merging this PR!

seanaery:
@jrochkind You are correct -- both issues I cited (`Apply` button gone awry and [lack of support for grouped search results](https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/blob/main/app/presenters/blacklight_range_limit/facet_field_presenter.rb?ref_type=heads#L45-53)) are pre-existing and not introduced by your branch. I'm motivated to help make this plugin work for all ArcLight implementers (which relies upon grouped search results), so I will indeed pursue that issue separately.

I'll report back this week on how this branch works with a BL7 + Rails 7 + jsbundling-rails + esbuild + BS4 app.

jrochkind:
@seanaery In a separate branch, I have a proposal to completely redesign the input form, which would resolve one of your problems!  For the `collapse` one I'm afraid I can't help as I'm not familiar with that feature at all. 

I want to wait to get this merged before PR'ing though, to avoid confusion/conflict.  Hopefully soon!

jrochkind:
Thanks to https://github.com/projectblacklight/blacklight/pull/3371 we were able to add a CI run that is now green for BL edge/main branch, Rails 8 beta, and esbuild.

seanaery:
I tried out this branch in a second Blacklight application of ours, the [Duke Digital Repository](https://repository.duke.edu/). [Here's my commit on my test branch](https://gitlab.oit.duke.edu/ddr/ddr-public/-/commit/c28dccf6f0cade9874c81ec363af442870ed6448). This app uses:
* Blacklight 7
* Bootstrap 4
* JSBundling-Rails w/ESBuild
* No importmaps

Things mostly work as advertised.
Screenshot 2024-10-30 at 2 39 51 PM

However -- there's a new bug I see happening, likely due to [this commit](https://github.com/jrochkind/blacklight_range_limit/commit/9c12a49adba76ca666ec46989921f67b32f66e0b) from a couple days ago. To corroborate, I updated the BL8/importmaps/Arclight app I successfully tested last week to the latest commit on `reboot_js_etc` and the same thing started happening.

At some browser widths, the `<canvas>` width starts at or drops to `0`, and as you resize the browser, it will correct itself to e.g., `447`, then return to `0`. I see that if you change the `display` property for the `<canvas>` from `block` to `inline-block` then this problem doesn't happen.

Here's a short 30sec video demonstrating the bug:
https://duke.box.com/s/9ld5fx1wis7u1rie6alzc4a6kz6mulct

jrochkind:
Good to hear it worked well for you and installed well on esbuild and Rails7. 

Thanks for reporting that problem and targetting it to that commit! I don't seem to reproduce that weird responsive error...   canvas.js doesn't really want you to set styles on the canvas yourself directly...  

I think I will for now just revert that commit that attempts to make the canvas.js resize as the window size changes dynamically. I think there are just some buggy behaviors there, that unfortunately chart.js doesn't seem in a hurry to fix. https://github.com/chartjs/Chart.js/issues/11005

But I'm not sure how vital a requirement it really is. We could try playing around with it more before a final release but for this PR I will revert.

seanaery (approved):
Thanks for all the great work here. Very thorough. I have tested these changes out using two very different existing BL applications and -- with the latest changes -- it works really well:
- ArcLight 1.x / Blacklight 8 / Importmaps (no build) / Bootstrap 5 / Rails 7
- Blacklight 7 / JSBundling-Rails (w/esbuild) / Bootstrap 4 / Rails 7

I approve and am inclined to go ahead and merge this PR. Just checking in with @bess to see if you'd like to test/review this first. Sounds like this'll clear the way for some smaller follow-up PRs that'll precede a beta release, and there'll be more opportunities to test in beta.

.github/workflows/ruby.yml
```diff
@@ -16,23 +16,104 @@ on:
...
+
         include:
-          - ruby: "3.2"
-            rails_version: "7.1.4"
-          - ruby: "3.1"
-            rails_version: "7.0.8.4"
+          # BLACKLIGHT EDGE, can test with Rails 8 beta, importmap and esbuild
+          #
+
+          - rails_version: "8.0.0.beta1"
```
jcoyne, line 40: Not a blocker for this PR, but lets put 8.0.0.rc2 in here.

jcoyne (approved):

.github/workflows/ruby.yml
jrochkind: might as well put it in this PR, thanks!

PR #267, opened 2024-11-04
Redesign input form for applying limits

jrochkind:
From (before):

Screenshot 2024-11-04 at 10 44 30 AM

To (after): 

Screenshot 2024-11-04 at 11 05 44 AM

* A bit challenging to do this with bootstrap utility classes only, as we don't have our own CSS and if possible don't want to introduce it. But did it!

* Required  i18n labels for shorter labels, which I have auto-translated for now with [i18n-tasks](https://github.com/glebm/i18n-tasks)

## Why?

* The old one, in order to work on both Bootstrap 4 and 5, had double labels with one marked sr-only/visual-only
   * Which itself may have accessibility problems, as screen-readers may see double labels
   * And was [breaking for at least some contexts, hiding all labels](https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/merge_requests/380). Just overly complex. 

* On the smallest sidebar size, the old one didn't actually have enough space for input (also could depend on user-agent font-size etc)
    * Screenshot 2024-11-04 at 10 44 42 AM

* Even non-screen-reader users could probably use on-screen labels to explain these boxes

### DIFF
--- a/app/components/blacklight_range_limit/range_form_component.html.erb
+++ b/app/components/blacklight_range_limit/range_form_component.html.erb
@@ -4,12 +4,12 @@
   <div class="range-limit-input-group">
     <div class="d-flex justify-content-between align-items-end">
       <div class="d-flex flex-column mr-1 me-1">
-        <%= label_tag(begin_input_name, t("blacklight.range_limit.range_begin_short"), class: 'text-muted small') %>
+        <%= label_tag(begin_input_name, t("blacklight.range_limit.range_begin_short"), class: 'text-muted small mb-1') %>
         <%= number_field_tag(begin_input_name, begin_value_default, class: "form-control form-control-sm range_begin") %>
       </div>
 
       <div class="d-flex flex-column ml-1 ms-1">
-        <%= label_tag(end_input_name, t("blacklight.range_limit.range_end_short"), class: 'text-muted small') %>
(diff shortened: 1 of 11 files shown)


jrochkind:
Not sure why CI isn't running?

jrochkind:
Working on the super odd failures, which I'm having trouble reproducing locally, and can't figure out why they'd only apply to the CI jobs they do. 

```
1) Run through with javascript basically works
     Failure/Error: click_button "Apply limit"

     Selenium::WebDriver::Error::ElementClickInterceptedError:
       element click intercepted: Element <input type="submit" value="Apply limit" class="submit btn btn-sm btn-secondary" data-disable-with="Apply limit"> is not clickable at point (368, 442). Other element would receive the click: <section id="content" class="col-lg-9 order-last" aria-label="search results">...</section>

```

seanaery (approved):
👏 This is really great. I have tested it out on two very different BL apps and it renders nicely in both cases:
* BL7/Bootstrap 4 digital collections app
* BL8/Bootstrap 5 vanilla generated ArcLight app

I like how it uses both v4 & v5 Bootstrap utility classes so it works with both.

**BL8/BS5:**
Screenshot 2024-11-06 at 11 41 02 AM

**BL7/BS4:**
Screenshot 2024-11-06 at 1 13 33 PM

It passes aXe DevTools accessibility testing as well. When I was reviewing these changes for accessibility, I spotted a keyboard navigation accessibility issue (not unique to this PR) that I will write up as a new issue.

PR #268, opened 2024-11-04
Remove all deprecated methods, templates, and other behavior

jrochkind:
In prep for future 9.0 release, with new JS

- Remove totally unused deprecated helper methods
- remove deprecated views
- remove deprecated format_range_display_value helper
- remove unused deprecated class methods in BlacklightRangeLimit
- remove entirely now-unused RangeLimitHelper
- remove deprecated behavior
- remove now-unused legacy template
- remove deprecated now-unused method
- we no longer use deprecation gem, have no deprecations


blacklight_range_limit.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'blacklight', '>= 7.25.2', '< 9'
   s.add_dependency 'view_component', ">= 2.54", "< 4"
-  s.add_dependency 'deprecation'
+  #s.add_dependency 'deprecation'
```
seanaery, line 22: Can we remove this altogether instead of commenting out?

jrochkind: Sure. 

I sort of last-minute decided to remove it, between that and leaving it as a dependency but currently unused since we have no deprecations to report.

PR #269, opened 2024-11-04
make chart.js responsive on screen size change

jrochkind:
When you drag the window to a new size, and it changes size of sidebar facet column, we want the chart to resize accordingly. 

While chart.js is supposed to support this with a feature they call "responsive", it's in a confusing situation with either a bug or unclear behvaior, where others are having a lot of trouble doing it too. https://github.com/chartjs/Chart.js/issues/11005

But I think we did it. 

Made the aspect ratio of the chart configurable while we're at it -- before it was always the chart.js default, which was `2` (width 2x height). 

So, with this implementation @seanaery reported a problem at https://github.com/projectblacklight/blacklight_range_limit/pull/266#issuecomment-2448091143

@seanaery, I haven't been able to reproduce your problem -- but I'm using a demo app where the sidebar with facets can only appear at a few specific sizes, it jumps between them at breakpoints. It sounds like you were demo'ing in a non-default app that had a fluid sized sidebar instead maybe?  Is it possible something *else* in your customized dom/css was causing the problem?

While I don't love it, I could be open to applying the solution you came up with if it really does seem necessary for you, as long as it's clearly commented to explain why we're doing it (without fully understanding it!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ config.add_facet_field 'pub_date', label: 'Publication Year',
                chart_replaces_text: true,
                chart_segment_border_color: "rgba(0,0,0, 0.5)",
                chart_segment_bg_color: "#ccddcc",
+               chart_aspect_ratio: "2"
             }
         )
 ```
@@ -117,6 +118,7 @@ config.add_facet_field 'pub_date', label: 'Publication Year',
 * **chart_replaces_text**: Default true. If false, when the chart is loaded purely textual facets will still remain on-screen too.
 * **chart_segment_border_color** / **chart_segment_bg_color** :
   * Set colors for the edge and fill of the segment bars in the histogram.
+* chart_aspect_ratio: for chart.js, will fill available width then this determines size of chart. defaults to 2
 
 
 ## Javascript dependencies
--- a/app/assets/javascripts/blacklight-range-limit/index.js
+++ b/app/assets/javascripts/blacklight-range-limit/index.js
@@ -174,6 +174,10 @@ export default class BlacklightRangeLimit {
     this.chartCanvasElement = this.container.ownerDocument.createElement("canvas");
     this.chartCanvasElement.setAttribute("aria-hidden", "true"); // textual facets sr-only are alternative
     this.chartCanvasElement.classList.add("blacklight-range-limit-chart");
+    // We set inline-block for compatibility with container-fluid layouts, e.g. when
+    // Blacklight's config.full_width_layout = true
+    // See: https://github.com/projectblacklight/blacklight_range_limit/pull/269
+    this.chartCanvasElement.style.display = 'inline-block';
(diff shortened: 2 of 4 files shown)


seanaery:
Thanks @jrochkind -- I took a closer look today especially to try to identify whether my applications are doing anything out-of-the-ordinary that's causing this to not work correctly for me ([behavior demonstrated in this 30sec video](https://duke.box.com/s/9ld5fx1wis7u1rie6alzc4a6kz6mulct)).

It's true that at Duke we tend to use more fluid layouts, and I confirmed that does come into play. Blacklight defaults to a fixed-width layout (where e.g., the outer container gets a `max-width: 1140px`), _but_ BL also provides a config option to switch to fluid (see [full_width_layout](https://github.com/projectblacklight/blacklight/blob/main/lib/blacklight/configuration.rb#L47)). If you set `config.full_width_layout = true`, that toggles the `#main-container` element's class from `container` to `container-fluid`. And if that's `container-fluid`, it does not (yet) play nicely with the responsive changes in this branch.

I spun up a fresh ArcLight/BL8 app to see whether this branch displays the histogram out-of-the box, and it did. But when I configured it with `config.full_width_layout = true` and/or just changed `container` to `container-fluid` in my browser devtools, it broke.

As an aside, I looked at what'd happen if you keep BL's default fixed layout but change the breakpoints, i.e., change the Bootstrap `$grid-breakpoints` & `$container-max-widths` variables. Things still worked fine with those changes.

While I can't fully explain exactly _why_ this fixes the fluid layout issue, I can confirm that a viable workaround is to change the `<canvas>` element from `display: block` to `display: inline-block`. This _could_ be done in an implementing app via hacky CSS like:
```
canvas.blacklight-range-limit-chart {
  display: inline-block !important;
}
```

But is probably more ideally handled via the plugin's JS like this:
```
this.chartCanvasElement.style.display = 'inline-block';
```

I think the fact that Blacklight provides a configurable switch to a fluid layout sways me toward wanting to see this accommodated by the plugin (i.e., not put the burden on the implementer).

jrochkind:
@seanaery while I don't love it without understanding it, and am worried what if it messes up some OTHER situation -- so be it. 

Are you interested in making a commit to this branch that does what you want, with the line being commented with a URL to this issue?  

When I have a moment, I'm going to try to focus on the "universal design" solution to the range interval facet value list.

seanaery:
I made that `display-inline` commit and pushed the branch back up. Relatedly, looking at the https://codepen.io/ouqu5/pen/ZEjEPqm demo cited in https://github.com/chartjs/Chart.js/issues/11005 -- applying `display: inline-block` to the `canvas` there fixes that issue.

jrochkind:
Awesome, thank you, and thanks for the comment. 

Let's go ahead and merge like this then?

Perhaps you want to tell them on the chart.js thread of your findings!

PR #273, opened 2024-11-11
Enable the plugin to work with grouped search results (e.g., ArcLight…

### LINKED ISSUE #272: Grouped search results views break the range facet display (ArcLight incompatibility)
### Overview
When search results are grouped (a feature essential to [ArcLight](https://github.com/projectblacklight/arclight)), the range limit facet will show zero results for the selected range, and fail to render a histogram.
Screenshot 2024-11-07 at 4 24 58 PM

This is a long-standing issue that pre-dates v9.x of `blacklight_range_limit`; it does not appear that the plugin has ever accounted for grouped search results.

### Steps to Reproduce
This would be true of any Blacklight app that uses grouped search results. ArcLight certainly does; TBD which other BL-based applications are impacted.
* In a new ArcLight app w/`blacklight_range_limit` added, perform a blank search
* Enter any date range in the facet; click `Apply` button
* Plugin works correctly here (shows correct hits for selected range and displays histogram)
* Toggle results to `Grouped by collection`
* Plugin will show `0` hits for the selected range (and fail to show the histogram)

### Why This Matters
The majority of current ArcLight implementations have added the `blacklight_range_limit` gem and have had to make various local fixes to get it to work correctly. Grouping by collection is essential and the community even supports making that the default view of search results (see https://github.com/projectblacklight/arclight/issues/1488). It would be quite helpful to the community if this plugin would just work out-of-the-box for ArcLight without requiring local modifications.

### Some Technical Details
The `BlacklightRangeLimit::FacetFieldPresenter` currently expects the Solr response for search results to be a `Blacklight::Solr::Response` and for the total hit count to be in `response.total`, aka `response['numFound']` ([see code](https://github.com/projectblacklight/blacklight_range_limit/blob/main/app/presenters/blacklight_range_limit/facet_field_presenter.rb#L38-L42)).

E.g., for a query yielding 510 docs:

```json
{
  "responseHeader": {...},
  "response": {
    "numFound": 510,
    "start": 0,
    "numFoundExact": true,
    "docs": [...]
  },
  "facet_counts": {...},
  "highlighting": {...},
  "stats": {...}
}
```

But when search results are grouped, the Solr response does not use `numFound`. The total doc count will instead be in `['grouped']['_root_']['matches']`, where `_root_` is the applied `group.field`, e.g. :

```json
{
  "responseHeader": {...},
  "grouped": {
    "_root_": {
      "matches": 510,
      "ngroups": 8,
      "groups": [...]
    }
  },
  "facet_counts": {...},
  "highlighting": {...},
  "stats": {...}
}
```

seanaery:
…). Closes #272.

* Note that arclight 1.x uses `'_root_'` as the `group.field` (indexes collection-component relationship using nested Solr docs), and `arclight 0.x` used `'collection_ssi'`; this solution will work with either.
* Accommodates all known grouping strategies enabled by Blacklight

Screenshot 2024-11-11 at 9 11 05 AM

### DIFF
--- a/app/presenters/blacklight_range_limit/facet_field_presenter.rb
+++ b/app/presenters/blacklight_range_limit/facet_field_presenter.rb
@@ -38,7 +38,7 @@ module BlacklightRangeLimit
     def selected_range_facet_item
       return unless selected_range
 
-      Blacklight::Solr::Response::Facets::FacetItem.new(value: selected_range, hits: response.total)
+      Blacklight::Solr::Response::Facets::FacetItem.new(value: selected_range, hits: selected_range_hits)
     end
 
     def missing_facet_item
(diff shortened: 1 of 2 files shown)


jrochkind:
I don't know how grouping works or how Blacklight deals with it, but the code change seems straightforward enough and the tests seem good. Thanks! Will approve and merge!

PR #274, opened 2024-11-12
Textual facets in collapsible on screen

jrochkind:
A "universal design" approach to the need for textual facets corresponding to the chart for low-vision or screen-reader-using users.

Putting all the links in a visually-hidden didn't work, becuase you can't actually put focasable things in visually-hidden without problems or more weirdness, so trying this. 

Put them on all screens, but inside a collapsible header, to avoid noise for those who don't want them.  

Rearranged some DOM and configuration and made some selectors more flexible in the process. 

Still needs i18n on label, and some other cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,8 @@ config.add_facet_field 'pub_date', label: 'Publication Year',
                assumed_boundaries: [1100, Time.now.year + 2],
                segments: true,
                chart_js: true,
-               chart_replaces_text: true,
+               textual_facets: true,
+               textual_facets_collapsible: true,
                chart_segment_border_color: "rgba(0,0,0, 0.5)",
                chart_segment_bg_color: "#ccddcc",
                chart_aspect_ratio: "2"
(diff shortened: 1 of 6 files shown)


seanaery:
That looks great to me, @jrochkind . Bonus points for making it configurable. I like the way the `details` & `summary` tags work here, and it's nicely responsive and keyboard navigable. As you noted, an `i18n` key for the label is needed so implementers have an easy hook to customize that. Here are screenshots running this branch in my demo app.

Screenshot 2024-11-12 at 2 20 01 PM
Screenshot 2024-11-12 at 2 20 09 PM

jrochkind:
OK, i18n added, initial non-English translations provided by Google Translate via i18n-tasks.

jrochkind:
@seanaery What do you think about that heading "Range List"?

I'm not sure anyone will have any idea what it means...  but having trouble coming up with anything. 

Colleague @eddierubeiz suggests "Number of results in each range", or "Results in each range", or maybe "Results in each interval"... 

I really don't know.

jrochkind:
Yeah, I think it would have to add custom JS to switch label for open/closed, which I was trying to avoid. 

Want to approve and merge what's here now?

seanaery (approved):

PR #275, opened 2024-11-12
Simply documented and tested config to activate range limit facet

jrochkind:
In the old days, you activated a range limit config with eg

    config.add_facet_field 'pub_date_si', label: 'Publication Date Sort', range: true

At some time that was changed to use a `default_range_config` method mixed into CatalogController...

    config.add_facet_field 'pub_date_si', label: 'Publication Date Sort', **default_range_config

The former simpler way of doing it now works again. And you can override other default config simply by adding the config you want to override:
    
    config.add_facet_field 'pub_date_si', label: 'Publication Date Sort', range: true, range_config: { textual_facets: false }

Changed docs and tests to go back to simpler original way. Also will include specific docs for over-riding Components. 

For now, the mixed-in `default_range_config` method is still available in CatalogControllers.. we could remove it, it would be backwards incompat with any existing config, but it is a major version, any thoughts? I'd just leave it in for now prob... maybe deprecate?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ package.json-based use will additionally need to point to the matching unreleaes
 You have at least one solr field you want to display as a range limit, that's why you've installed this plugin. In your CatalogController, the facet configuration should look like:
 
 ```ruby
-config.add_facet_field 'pub_date', label: 'Publication Year', **default_range_config
+config.add_facet_field 'pub_date', label: 'Publication Year', range: true
 ```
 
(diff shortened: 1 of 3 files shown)


jrochkind:
not sure why CI isn't running, but I built on top of #274 to avoid conflict, so will rebase and make Ready when that one is resolved.

seanaery:
This looks like a good change to me, making this syntax what's recommended in the README. Nice to see that works again, as it feels cleaner/simpler. No strong opinion on whether to drop, keep, or deprecate support for `**default_range_config`.

Issue #237, opened 2023-07-05
Blacklight 8, empty range facet shown even for zero results page


jrochkind:
I wanted to reproduce this on a fresh empty app, but I've been unable to create a fresh empty Blacklight 8 app with `blacklight_range_limit`, per #236 

But I do somehow have a branch of my actual app working with `blacklight_range_limit 8.3.0`, `blacklight 8.0.1`, and `rails 7.0.5.1`.  (I am not sure how I got this to work, and have not been able to reproduce in a new blank app!)

Here I have only one problem.  In cases of zero results, when the Blacklight "zero results" message is shown, with Blacklight 7, you get _no_ facets in sidebar -- this is expected. 

In Blacklight 8, my single `blacklight_range_limit` facet is still showing up in sidebar on "zero results" page, and with facet body content that does not make any sense. 

I _believe_ that this is not about anything otherwise custom to my app, and could be reproducible in a stock app -- if anyone can make one!  But these screenshots do show CSS customizations, and the custom "other search tools" static content we've added to our sidebar. 

## Blacklight 7

As expected, no facet content in "zero results" page. ("Other search tools" static content in sidebar is an unrelated customization). As there are no available facets to display at all, we don't even get the "Limit your search" facet section header. 

![Screen Shot 2023-07-05 at 1 30 10 PM](https://github.com/projectblacklight/blacklight_range_limit/assets/149304/70f0c5aa-80b0-4fcf-b908-f41df9f24854)

## Blacklight 8

"Date" is our range facet -- it is showing up in sidebar even in "zero results" page, but with a body that doesn't make any sense. This should not be showing up in a zero results page, as it was not in BL 7. 

![Screen Shot 2023-07-05 at 1 32 34 PM](https://github.com/projectblacklight/blacklight_range_limit/assets/149304/63b38ad0-a250-45d1-ab02-4e889e793002)

jrochkind:
This is related to https://github.com/projectblacklight/blacklight/issues/3054 , but is also it's own problem. 

If the problem were solved so blacklight_range_limit were not rendering a box for an "empty" result set, then you would still get "Limit your search" header because of https://github.com/projectblacklight/blacklight/issues/3054 .  But there's a separate pre-requisite problem here, to make it stop rendering the nonsensical facet box for empty results.

jrochkind:
I think I have a solution, which would be implementing `BlacklightRangeLimit::RangeFacetComponent#render?` as so:

```ruby
[...]

PR #276, opened 2024-11-12
Display properly when there is no data

jrochkind:
- keep chart wrapper with explicit aspect-ratio from taking up space when there is no chart
- avoid JS error when there is no data available

## Before

Screenshot 2024-11-12 at 12 25 05 PM

Plus a JS error. 

## After

Screenshot 2024-11-12 at 12 29 34 PM

With no JS error.


seanaery:
That's a nice change. It also makes #237 not stand out so much.

PR #278, opened 2024-11-13
Support open-ended ranges

### LINKED ISSUE #235: If you omit the start or end date, the search is not filtered.
Start with a standard-issue test site with recent versions of `blacklight` and `blacklight_range_limit`:

```
gem 'blacklight', "= 7.32.0"
gem "blacklight_range_limit", "=8.2.3"
```

### Scenario
- Visit http://localhost:3000/ ;
- Open up the range facet on the left;
- Note that the dates in the test data range between 1941 and 2008;
All fields
- In the facet form above, set the start date to blank, and set the end date to 2000 - or just visit http://localhost:3000/?range[pub_date_ssim][begin]=&range[pub_date_ssim][end]=2000 .

### Expected behavior
- I expect the range facet dropdown to stay open;
- I expect a histogram showing results between 1941 and 2000;
- I expect to see results filtered according to those dates.

I believe BLR showed the expected behavior **until version 8.1.0, at the earliest.**

### Actual behavior
- I'm redirected a page that looks identical to the root of the test website;
- No results are shown;
- The range facet dropdown is closed;
- Opening it, the range dates are reset to 1941 and 2008.

I believe BLR started showing the current behavior **starting with version 8.2.2, at the latest.**

jrochkind:
Fixes #235

Relies on behavior that will be in upcoming Blacklight release from https://github.com/projectblacklight/blacklight/pull/3443 -- but polyfills this behavior in, kind of hackily. (It would have been even hackier to try to keep things no-op on old Blacklights without the fix I think!). Assume BL fix will be in next BL 8.x release, since it's already merged. 

Include some new browser specs meant to try to prevent regression, as this was a breakage tests didn't fix caused by a major refactor previous. 

I started work on this before noticing @taylor-steve's #258, but once I did notice reivewed his code and saw a couple things I had missed actuallyand ideas for improving mine, and borrowed his tests. Thanks @taylor-steve !

- Polyfill Blacklight endless range behavior
- Support one-ended ranges
- add browser specs for open-ended ranges

### DIFF
--- a/spec/features/run_through_spec.rb
+++ b/spec/features/run_through_spec.rb
@@ -61,6 +61,50 @@ describe 'Run through with javascript', js: true do
     end
   end
 
+  context "open-ended range" do
+    it "can search" do
+      visit search_catalog_path
+
+      click_button 'Publication Date Sort'
+
+      within ".facet-limit.blacklight-pub_date_si" do
+        find("input#range_pub_date_si_begin").set("")
+        find("input#range_pub_date_si_end").set(end_range)
+        click_button "Apply limit"
+      end
+
+      expect(page).to have_css(".applied-filter", text: /Publication Date Sort +to #{end_range}/)
+      expect(page).not_to have_text("No entries found")
+      expect(page).to have_css(".document")
+
+      within ".facet-limit.blacklight-pub_date_si" do
+        # expect expandable limits
+        find("summary", text: "Range List").click
+        expect(page).to have_css("details ul.facet-values li")
+      end
+    end
+  end
+
+  context "submitted with empty boundaries" do
+    it "does not apply filter" do
+      visit search_catalog_path
+
+      click_button 'Publication Date Sort'
+
+      within ".facet-limit.blacklight-pub_date_si" do
+        find("input#range_pub_date_si_begin").set("")
+        find("input#range_pub_date_si_end").set("")
+        click_button "Apply limit"
+      end
+      expect(page).not_to have_css(".applied-filter")
+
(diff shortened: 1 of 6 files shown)


spec/features/run_through_spec.rb
```diff
@@ -61,6 +61,57 @@
...
+
+      within ".facet-limit.blacklight-pub_date_si" do
+        # expect expandable limits
+        find("summary", text: "Range List").click
+        expect(page).to have_css("details ul.facet-values li")
+      end
+    end
+  end
+
+  context "submitted with empty boundaries" do
```
seanaery, line 90: I found this spec a little counterintuitive -- maybe I'm just misunderstanding what is meant by "empty boundaries"? Looks like the test fills in lower and upper boundaries (begin 1900, end 2100) and applies them, so I'm confused why this would result in no `.applied-filter` nor `.selected` elements present.

seanaery (commented):
Everything here works really well--nice job on this. Just added one question I have about one of the feature specs.

spec/features/run_through_spec.rb
jrochkind: On line 65, ` let(:start_range) { "nil"}`, so it does not enter anything into the first box. 

Perhaps that is a confusing way to structure things yeah, there's no need for that, I'll alter it!

jrochkind: I'm finding some other odd things in there that weren't what I intended, thanks for bringing to my attention! That was indeed a confusing way to write it. Will change commit.

seanaery (approved):

PR #280, opened 2024-11-13
Don't show chart or textual facets when only one facet bucket,

### LINKED ISSUE #224: Chart should not render for single date result
In BlacklightRangeLimit 6.3.3, a single date result is shown as a linked facet and the chart is not rendered, as shown here:
BRL633singleresult

But this does not appear to be the default behavior in version 8.2.4:

BRL824singleresult

jrochkind:
It's not useful to show a histogram with only one bucket, or a facet list with only one value. 

The expected case is when our limit is only one year. 

Closes #224

- move extract data into separate step, so we can use data in deciding to show text facets
- Don't show chart or textual facets when only one facet bucket, not useful


seanaery (approved):
I tried out this branch and it looks good. This is a welcome improvement.

PR #281, opened 2024-11-14
Don't render the range limit facet on zero results pages; fixes #237.

### LINKED ISSUE #237: Blacklight 8, empty range facet shown even for zero results page

seanaery:
- reorganize impacted existing specs so they have the data they need to render the range_facet_component

This implements @jrochkind 's idea (and annotation) described in #237 with a minor adjustment to ensure the component's predicate method `render?` returns a boolean `true` or `false`. It adds two new specs verifying that the facet doesn't render, and moves a few existing component specs so they continue to pass.

### DIFF
--- a/app/components/blacklight_range_limit/range_facet_component.rb
+++ b/app/components/blacklight_range_limit/range_facet_component.rb
@@ -10,12 +10,22 @@ module BlacklightRangeLimit
       @classes = classes
     end
 
+    # Don't render if we have no values at all -- most commonly on a zero results page.
+    # Normally we'll have at least a min and a max (of values in result set, solr returns),
+    # OR a count of objects missing a value -- if we don't have ANY of that, there is literally
+    # nothing we can display, and we're probably in a zero results situation.
+    def render?
+      (@facet_field.min.present? && @facet_field.max.present?) ||
+        @facet_field.missing_facet_item.present?
+    end
+
     def range_config
       @facet_field.range_config
     end
 
     def range_limit_url(options = {})
-      helpers.main_app.url_for(@facet_field.search_state.to_h.merge(range_field: @facet_field.key, action: 'range_limit').merge(options))
+      helpers.main_app.url_for(@facet_field.search_state.to_h.merge(range_field: @facet_field.key,
+                                                                    action: 'range_limit').merge(options))
     end
 
     def uses_distribution?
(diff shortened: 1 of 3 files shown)


seanaery:
Interesting -- I can't think of any cases where we'd encounter that scenario, but I'm also not 100% certain. The way I think of it is, the facet UI element exists for three purposes: 1) narrow down the current result set, 2) show how many docs in the current result set match the currently selected value(s) for the facet, 3) remove a currently selected value w/`X`.

If the current result set includes no documents that have a value in the range field, then no matter what values a user would enter into the form, it won't get them any documents, so it seems safe not to render the form. 

The facet itself _does_ still render (without the chart & form) in this scenario where there's no `min` & `max`:
* User clicks `[Missing]` in the facet
* Result set is only documents that _don't have a value_ in the range field
* Facet still renders, so user see they have selected `[Missing]`, and can `X` that selection to remove the filter

PR #290, opened 2024-11-21
For open-ended limit, ensure both limit and data min/max are taken account of in calculating range bucket boundaries

jrochkind:
In our app we use a multi-valued field for our range limit. In that case, when entering an open-ended limit, range limit could accidentally calculate boundaries for the buckets that exceeded the specified single-end boundary. Because it was ignoring the boundaries unless both ends were present.

Fixed that logic, with tests.


seanaery (approved):
👏 I tried this out in an ArcLight app (archival description often uses a wide range of year values on a single document!). This worked great and is a welcome improvement.

PR #292, opened 2024-11-25
Make displaying the [Missing] link/value configurable (default true).…

### LINKED ISSUE #288: The [Missing] value/link may be too prominent -- create option to suppress?
Example screenshot from a BL8 app with configured chart color; note `[Missing]` under `Range List`:
Screenshot 2024-11-21 at 1 21 35 PM

From `#blacklight` Slack discussion 11/21/2024 -- thoughts for consideration:

@seanaery :
> The `[Missing]` value is prominent and feels noisy especially alongside `Range List` . In some apps we have hidden it with CSS. Could there simply be a config option added to omit displaying it? [...] And/or, might `[Missing]` be moved into the list that gets revealed upon clicking `Range List`? Maybe this'd get tricky.

@jrochkind :
> Missing — we here kind of believe it’s important to let users know there items that don’t have dates and aren’t showing up in teh chart or the range buckets at all, and won’t be included if you apply any date limits at all.
So that’s why I’m not sure about putting it inside a disclosure I don’t expect most to open.
But I’m not opposed to a config to turn off missing (it could probably actually effect the query to solr to simply not ask for the “missing” results).  although I’d personally recommend against using it.

@seanaery :
> [I] will plan to PR something for `[Missing]` this week that will only affect the display, not the actual query logic.

@jrochkind :
> I’d really suggest you look at effecting query logic instead!  Dont’ ask Solr for something we dont’ want to use, and then the display I’d guess already won’t display it if Solr does not offer it.

seanaery:
… Closes #288.

### DIFF
--- a/app/components/blacklight_range_limit/range_facet_component.html.erb
+++ b/app/components/blacklight_range_limit/range_facet_component.html.erb
@@ -41,7 +41,7 @@
           </div>
         <% end %>
 
-        <% if @facet_field.missing_facet_item && !request.xhr? && uses_distribution? && range_config[:show_missing_link] %>
+        <% if range_config[:show_missing_link] && @facet_field.missing_facet_item && !request.xhr? && uses_distribution? %>
           <%= render BlacklightRangeLimit::RangeSegmentsComponent.new(facet_field: @facet_field, facet_items: [@facet_field.missing_facet_item], classes: ['missing', 'mt-3']) %>
         <% end %>
       <% end %>
--- a/app/components/blacklight_range_limit/range_facet_component.html.erb
+++ b/app/components/blacklight_range_limit/range_facet_component.html.erb
@@ -41,7 +41,7 @@
           </div>
         <% end %>
 
-        <% if @facet_field.missing_facet_item && !request.xhr? && uses_distribution? %>
+        <% if @facet_field.missing_facet_item && !request.xhr? && uses_distribution? && range_config[:show_missing_link] %>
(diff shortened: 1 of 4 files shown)


app/components/blacklight_range_limit/range_facet_component.html.erb
```diff
@@ -41,7 +41,7 @@
           </div>
         <% end %>
 
-        <% if @facet_field.missing_facet_item && !request.xhr? && uses_distribution? %>
+        <% if @facet_field.missing_facet_item && !request.xhr? && uses_distribution? && range_config[:show_missing_link] %>
```
jrochkind, line 44: I'd put the config check first.

jrochkind:
I'm still really interested in trying to do this at the Solr request level instead.  It's putting load on solr that doesn't have to be put if you're not going ot use it. 

Let me see if I can find time this week to see if I can easily do that.

jrochkind:
Ah, I see 'missing' is currently coming from Stats component, not a separate facet added. And we need stats component anyway. So ok fine!

seanaery:
@jrochkind Good idea about switching the order of conditions checked before rendering the link -- thanks for that.

Yep, there's an existing test in the `basically works` block higher up in `run_through_spec.rb` that verifies the default missing item display (See  `expect(page).to have_link '[Missing]'`).

As for the load on Solr, yes, I came to that same conclusion after investigating further... it's just one request that we need anyway to get the `min` & `max` stats. Thanks!

jrochkind:
I still think in most cases it's bad UX to not tell people about search result items that will be excluded from any date range limit!  But I'm not the boss.

PR #294, opened 2024-11-26
Clamp values between config'd min and max

### LINKED ISSUE #289: Entering extremely large or small integers in the form results in an error
Should we impose min/max value constraints to prevent extremely large or small integers from even being submitted? Solr integer fields are 32-bit signed integers; as such the valid range is from `-2,147,483,648` to `2,147,483,647`.

If you enter a value outside those boundaries into the form you'll get a `500 Internal Server Error` with `Blacklight::Exceptions::InvalidRequest (RSolr::Error::Http - 400 Bad Request)` / `Error: Invalid Number: ### for field`.

jrochkind:
To me the main reason to do this is to avoid the possibility of sending something to solr that results in a 500 error. i hate these in my logs, eg created by bots trying random query strings to check for vulnerabilities -- which is why we need a check server-side too, not just client-side. 

- range values clamped by config'd min and max, to avoid solr error
- use html5 min/max on input field for config'd min/max

Closes #289

This PR does not yet have tests. Ideally I think it would have some for server-side (I'm not concerned with testing client-side min/max), but there were no existing tests for the builder, so figuring out the way to scaffold the test si annoying. 

@seanaery are you interested in adding a test to this?

### DIFF
--- a/app/components/blacklight_range_limit/range_form_component.html.erb
+++ b/app/components/blacklight_range_limit/range_form_component.html.erb
@@ -5,12 +5,22 @@
     <div class="d-flex justify-content-between align-items-end">
       <div class="d-flex flex-column mr-1 me-1">
         <%= label_tag(begin_input_name, t("blacklight.range_limit.range_begin_short"), class: 'text-muted small mb-1') %>
-        <%= number_field_tag(begin_input_name, begin_value_default, class: "form-control form-control-sm range_begin") %>
+        <%= number_field_tag(begin_input_name,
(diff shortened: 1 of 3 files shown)


seanaery:
This looks like a great approach to me. I'll be glad to add some tests and will plan to do that this afternoon...

jrochkind:
I intentionally didn't doc the new config vars, cause I don't think anyone should really use them, but they're there just in case for the deep divers.

seanaery (approved):

PR #298, opened 2024-11-27
hacky way to keep unpermitted params out of text range facet  links

### LINKED ISSUE #296: Clicking on a textual "Range List" facet does not meet Blacklight controller StrongParams requirements
So produces a log warning under default config, but an exception if Rails `config.action_controller.action_on_unpermitted_parameters == :raise`

See for comparison: https://github.com/projectblacklight/blacklight_range_limit/issues/196

Have to diagnose what's going on in this case -- but comparing to the built-in Sort select widget and see how it manages to get a list of "all current search params" should be a guide to solution. 

I think a browser test with raise config for click on link is appropriate regression test.

jrochkind:
It seems like there should be a better way -- I thought Blacklight search_state was only including allow-listed params in the first place but apparently not?  Not sure what a better way would be or if it would require Blacklight architectural changes, but this works for now. Regression test included that was red without this change.

Closes #296

### DIFF
--- a/app/presenters/blacklight_range_limit/facet_item_presenter.rb
+++ b/app/presenters/blacklight_range_limit/facet_item_presenter.rb
@@ -8,6 +8,14 @@ module BlacklightRangeLimit
       label_for_range || super
     end
 
+    # Very hacky way to keep params used for ajax query for segments out
+    # of our facet links. There oughta be a better way?
+    #
+    # https://github.com/projectblacklight/blacklight_range_limit/issues/296
+    def href(path_options = {})
+      super(path_options.merge(range_end: nil, range_field: nil, range_start: nil))
+    end
+
     private
 
     def label_for_range
--- a/app/presenters/blacklight_range_limit/facet_item_presenter.rb
+++ b/app/presenters/blacklight_range_limit/facet_item_presenter.rb
@@ -9,7 +9,7 @@ module BlacklightRangeLimit
     end
 
     # Very hacky way to keep params used for ajax query for segments out
-    # of our facet links. There oughta be a better way?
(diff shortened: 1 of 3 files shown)


jrochkind:
Not sure why it's failing tests and it may be unrelated to the change ("warning: conflicting chdir during another chdir block" what?)

but I realized i want to look into this more to understand Blacklight's relevant architecture better first too.

jrochkind:
OK, I learned more about what's going on. Blacklight SearchState _does_ have an allow-list, such that un-mentioned params are ignored. 

But we had to ADD our special range params into the allow-list, to get them available for use in the builder where we needed them. 

https://github.com/projectblacklight/blacklight_range_limit/blob/c11590f22646ce256582edb3757251205a824586/lib/blacklight_range_limit/controller_override.rb#L11-L19

But we want them ONLY in the #range_limit action -- we don't want them forwarded on when we generate facet links to typical catalog action. 

I actually can't see any better way to get that to happen then with the overide in FaceItemPresenter. 

BUT, two improvements as a result of this investigation:

* We can move the per-action addition of our `#range_limit`-specific allowed search params into that method to effect _only_ that action, and _not_ effect #index or any actions in the host controller, much more careful!

* We can re-use the constant that already held the keys of those params, when we strip them out in FacetItemPresenter, to keep the list DRY.

jrochkind:
@seanaery after this one, I'll release a beta2 -- and hope to be done!

seanaery:
This looks to me like a reasonable way to address the issue, even if it is imperfect. Thanks for figuring it out. The test is helpful and I like that this isolates the params the plugin needs to permit within the `range_limit` action.

Issue #300, opened 2024-12-02
Automatically reverse out of order boundaries


jrochkind:
Right now if you enter begin and end in reverse order, say `begin: 1950`, `end: 1900` -- an exception gets raised, is not caught, resulting in a 500 in an app. 

It would be nice if these were just automatically reversed to do the only thing that can be done with them instead. 

I started to look at this, but there were a lot of layers, and i wasn't sure of a good way to do it. So prob not to be included in 9.0.0. 

In my own app, I detect and reverse them with a `before_action`, but i'm not sure this is actually the best or right way when we can change code at layers a consuming app doesn't have access to. 

https://github.com/sciencehistory/scihist_digicoll/blob/f76ac4e22959fa344226546bb7cd09bda874612c/app/controllers/catalog_controller.rb#L10C18-L10C51

https://github.com/sciencehistory/scihist_digicoll/blob/f76ac4e22959fa344226546bb7cd09bda874612c/app/controllers/catalog_controller.rb#L669-L682

Here is a controller test that could test for it....

```ruby
require 'spec_helper'

RSpec.describe CatalogController, type: :controller do
  include Devise::Test::ControllerHelpers

  let(:range_facet_field) { "pub_date_si" }

  describe "range value out of order" do
    render_views

    let(:out_of_order_range_params) do
      {
        range: {
          year_facet_isim: {
            begin: "2010",
            end: "2000"
          }
        }
      }
    end

    it "treats as if ordered properly" do
      get :index, params: out_of_order_range_params
      expect(response.status).to eq(200)
      expect(response.body).to include("2000 to 2010")
    end
  end
end
```

PR #301, opened 2024-12-02
regression tests for existing kinds of bad params currently handled okay

jrochkind:
But in past versions of bl_range_limit i know from my own app resulted in uncaught exception 500s


jrochkind:
Does not include #300 because I exceeded my time budget trying to resolve it.

jrochkind:
Aha, one example is actually NOT handled okay in BL7, although is in BL8. 

Trying to figure out what's up.

PR #305, opened 2024-12-04
bad query params to #range_limit action should not result in uncaught exception

jrochkind:
Note that raising these specific excpetions will be automatically turned by rails into BadRequest => http 400, and NotFound => http 404 response.

The basic goal is that there should be no URL you can construct that will reuslt in an uncaught exception. 

Because I hate it when my exception monitor alerts me for things that were random non-working URLs that some bot looking for vulnerabilities or malfunctioning came up with. They didn't find vulnerabilities, but they did find something that caused the logic to raise unexpectedly. 

These are all cases seen in the exception monitor in my actual deployed production app; I've been catching/ignoring them locally.

### DIFF
--- a/lib/blacklight_range_limit/range_limit_builder.rb
+++ b/lib/blacklight_range_limit/range_limit_builder.rb
@@ -49,7 +49,8 @@ module BlacklightRangeLimit
     def fetch_specific_range_limit(solr_params)
       field_key = blacklight_params[:range_field] # what field to fetch for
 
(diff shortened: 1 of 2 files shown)


jrochkind:
Sadly too late for beta2!

jrochkind:
Aha, requires some extra logic to work in BL 7.x too, forthcoming.

seanaery (approved):

PR #308, opened 2025-01-07
Just a test of current CI: empty commit to test CI

jrochkind:


jrochkind:
have succesfully reproduced locally. not yet debugged or diagnosed, tricky to figure out what's up with propshaft.

jrochkind:
OK in generated .internal_test_app application.js there is indeed a duplicate:

```
import githubAutoCompleteElement from "@github/auto-complete-element";
import githubAutoCompleteElement from "@github/auto-complete-element"
```

So have to figure out what's generating that and fix it! Maybe some version of BL 7.x started adding something not previously added.

jrochkind:
Based on run history, seems build worked for this cell with BL 8.6.1, but broke with BL 8.7.0.  (it actually failed in a PR we merged anyway recently, not sure why we did that!). 

Haven't yet reproduced locally.

jrochkind:
OK, I think our generator for BL 8 fixup is no longer necessary in BL 8.7.0 and in fact causes problems. Good that BL 8.7.0 changed to make our extra generator fixup not necessary, but, phew! 

Will PR separately.

seanaery:
@jrochkind Thanks so much for tracking all that down. The test matrix all passed in the original PR #305 which was issued Dec 3, so that would've been using BL 8.6.1. Once the PR was merged to `main` on Dec 6, BL8.7.0 had been recently cut (Dec 4), so it adds up that a change in BL 8.7.0 is the culprit.

jrochkind:
Fixed in #309

PR #309, opened 2025-01-07
Test App fixup for BL8 only needed prior to 8.7.0

jrochkind:
And actually causes problems in 8.7.0, trying to do something that was already done. See #308


seanaery (approved):
Good detective work here. Thanks for fixing.

Issue #312, opened 2025-02-06
How do I render range limiting form fields as part of an advanced search form in the newer version of this gem.


dkinzer:
I'm trying to upgrade to the latest version of this gem (v9.0.0). To my dismay, as far as I can tell, the `render_range_input` helper method has been completely removed from the gem without providing a substitute way to render input fields.  The components that are meant to replace the helper method only render the input fields in the context of  the range limit form.  But, if I want to embed a range field inside an advanced search form I cannot (because then I will have a form within a form situation).

I think this helper method should be returned, or maybe input field only component can be used.

jrochkind:
Hi here is the commit where that went away, by me in preparing 9.0 release:

e478eec2b5bde86ec29ad1f6fd772d577d22816f

However, it had already been deprecated as going away in 9.x, 2.5 years ago in May 2022 and version 8.2, here:

32d073f9

That was part of refactoring to work with the new "view component" architecture, to get it to work with refactored blacklight. I am not totally sure the whether that deprecated helper method *worked*  in what versions -- I'm curious if you were using in 8.2.0+, or if you are migrating from an older version?

I did not totally understand the great view component refactor at the time, and I actually did think of this use case I'm afraid, and tried to bring it to the attention of those doing the refactoring; I'm not sure if I succesfully got it across, but at the time, the answer I recall was "pull requests welcome". 

I think you are supposed to instead try to look at configuration for a view component configured and use that maybe?  

I _think_ there is a component that is not a form within a form, but just the field you need? Oh.... I see maybe there wasn't because it wasn't needed for internal use cases. 

I do understand your use case and feel your pain. 

I guess the first step is factoring out just the field as a separate ViewComponent, so you can re-use it?  Or is there another refactor that would give you what you need?

Once that's done, a helper method could be returned (it's I think a pre-requisite for adding a helper method), or there might be another better way to provide access to that Component itself? 

Are you interested in trying to do the PR? If so, I will commit to reviewing it and working with you if needed!

I'm sorry this use case was missed in the refactor; I spent around two solid months saving blacklight_range_limit to be able to work with Blacklight 8 and 9 at all, and don't use advanced search form myself, and there were no tests inside this project that ensured it. If we add it back, we should make sure to also add a test that would fail if it were broken or removed, and ensures the "api" used by a dependent project like blacklight_advanced_search remains working!

dkinzer:
[...]

PR #313, opened 2025-02-07
Add and use input field components for range form.

dkinzer:
REF: #312

### DIFF
--- a/app/components/blacklight_range_limit/range_form_component.html.erb
+++ b/app/components/blacklight_range_limit/range_form_component.html.erb
@@ -3,25 +3,21 @@
 
   <div class="range-limit-input-group">
     <div class="d-flex justify-content-between align-items-end">
-      <div class="d-flex flex-column mr-1 me-1">
-        <%= label_tag(begin_input_name, t("blacklight.range_limit.range_begin_short"), class: 'text-muted small mb-1') %>
-        <%= number_field_tag(begin_input_name,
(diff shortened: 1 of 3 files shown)


jrochkind:
Ah, I see you already have gotten to it. 

What do you think about adding a configuration field for the input_component, so local apps can change it and have it picked up by the advanced search plugin?

What tests should we add to ensure the hook you need in advanced search from remains without breaking, perhaps with comments warning that this is API used by external dependents?

jrochkind:
The test failure looks familiar, this is something that's been happening to other projects, perhaps updating Rails version to latest patch release would fix it, I forget.

seanaery:
Yes, the test failure is [this](https://github.com/projectblacklight/blacklight/issues/3493); it's happening in the Rails 7.0 cell and is due to an internal Rails bug that isn't going to get fixed given that 7.0 is no longer under active support. Workarounds are challenging and/or undesirable, e.g., pinning `"concurrent-ruby", "< 1.3.5"`. The solution at the Blacklight core level was to remove any Rails versions under 7.1 from the CI matrix. Seems like we should we do the same here in `blacklight_range_limit`.

jrochkind:
Thanks! 

I think pinning concurrent-ruby to < 1.3.5 for test matrixes with Rails less than 7.1 is the way to go. 

I'll try to do a PR to do that separately.

PR #314, opened 2025-02-11
old Rails require old concurrent-ruby when running in CI

jrochkind:
Realized problem in #313

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,9 @@ else
       gem 'sass-rails', '~> 5.0'
       gem 'sprockets', '~> 3.7'
       gem 'thor', '~> 0.20'
+    when /^7.0/, /^7.1/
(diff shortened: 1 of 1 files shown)


jrochkind:
Hm. It may actually not be possible to workaround. 

The error happens during the engine cart build phase.  Not sure why we can't control what version fo concurrent-ruby is available there by Gemfile though. Havne't quite given up yet.

jrochkind:
engine cart man, gahh.

jrochkind:
OK maybe we do just have to stop testing EOL rails. sorry for those still on it.

seanaery:
A valiant effort trying to work around it - thanks for looking into it @jrochkind

jrochkind:
Not sure how we currently are green without this but apparently this PR is not necessary, perhaps the fix ended up in the large PR's to get to 9.0.

PR #315, opened 2025-03-28
Stop testing on rails 7.0 & ruby 3.1; both EOL.

seanaery:
Note this would unblock #313

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -24,8 +24,8 @@ jobs:
         #
         #   * Blacklight: 7, 8, or 9
         #   * Rails: 7 or 8
-        #   * sprockets or propshaft(with cssbundling-rails sass)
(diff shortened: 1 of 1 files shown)


jcoyne (approved):

.github/workflows/ruby.yml
```diff
@@ -42,29 +42,30 @@ jobs:
...
           - rails_version: "7.2.1"
             blacklight_version: "~> 8.0"
             ruby: "3.3"
             additional_name: "/ importmap-rails, propshaft"
             additional_engine_cart_rails_options: "-a propshaft --css=bootstrap"
+
           - rails_version: "7.2.1"
             blacklight_version: "~> 8.0"
-            ruby: 3.3
+            ruby: "3.3"
```
jcoyne, line 68: There's no need to quote this.

PR #316, opened 2025-06-19
Avoid using deprecated Blacklight class

jcoyne:


seanaery (approved):
I see, these classes were improved with a `Facets` namespace in https://github.com/projectblacklight/blacklight/pull/3614 -- so, this is a welcome change here 👍

PR #317, opened 2025-08-06
add more cases to CI, newer rails and blacklight versions

jrochkind:


jrochkind:
OK got it working, and the `git` one SHOULD be using view component 4... but weirdly when I run locally with the git edge one, it uses view_component 4 and _fails_, not sure how it passes here. 

Will add more code to force running view component 3 vs 4 in new versions, and ensure green.

seanaery:
Looks good to me, @jrochkind . Thanks for tackling this!

PR #319, opened 2025-08-07
Version bump 9.1.0

jrochkind:


jrochkind:
Good catch thank you very much, I forgot we had a JS package.

PR #321, opened 2025-10-07
avoid dependency conflict error by avoiding explicit json require in Gemfile

jrochkind:
`You have already activated json 2.15.0, but your Gemfile requires json 2.13.2` is what we were getting in CI runs. 

Comes about because we are trying to parse json in the gemfile to determine what dependencies we want to specify, for CI. 

Somehow we were activating a system json for use in the Gemfile production that was not hte same json Gemfile dependency resolution arrived at. 

Not sure what the workaround is, may have to totally change this system.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -34,19 +34,22 @@ jobs:
         # that'd be too much. Plus BL can't *generate* every one even though it
         # can be made to work with every one. We test with some good significant
         # representative samples.
+        #
+        # If 'blacklight_version' includes `://`, will be understood as github url,
+        # used for testing off blacklight main.
 
         include:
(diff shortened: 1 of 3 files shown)


jrochkind:
OK, we have to seriously un-clever-ize it -- if `://` is in `BLACKLIGHT_VERSION` we assuem it's a git url, and pass it asn argument `git: ENV['BLACKLIGHT_VERSION']`.  

No good way to stuff structured hash data in there, without dependencies, and you can't actually use dependencies in the Gemfile itself. 

If in the future we need to use a git branch, we'll have to figure that out then, good enough for now. 

@seanaery , interested in review? thank you!!

jcoyne (approved):

PR #325, opened 2025-11-04
Create a search builder for the range limit route

jcoyne:
This helps follow the pattern of having a class for each search builder rather than creating them dynamically.  It may also allow Blacklight to remove the SearchBuilder#append and #except methods

### DIFF
--- a/app/models/range_limit_search_builder.rb
+++ b/app/models/range_limit_search_builder.rb
@@ -1,0 +1,5 @@
+# Used for building the search for the range_limit route
+# You may completely override this class in the host application if you want to customize the search behavior.
+class RangeLimitSearchBuilder < SearchBuilder
+  include BlacklightRangeLimit::RangeLimitSearchBuilderBehavior
+end
--- a/lib/blacklight_range_limit.rb
+++ b/lib/blacklight_range_limit.rb
@@ -3,6 +3,7 @@ module BlacklightRangeLimit
   require 'blacklight_range_limit/facet_field_config_override'
   require 'blacklight_range_limit/range_limit_builder'
   require 'blacklight_range_limit/controller_override'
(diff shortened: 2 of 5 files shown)


jcoyne:
> Hi, so right now my app has local cusotmizations it wants to apply to search builders -- say authorization stuff.
> In Blacklight 9.beta8, my local app applies them in it's local SearchBuilder. I already knew that it was proposed it would have to apply them additionally in FacetSearchBuilder.
> With this proposed change, would it also have to apply them in a locally customized RangeLimitSearchBuilder? So all users of blacklight_range_limit would have to know that if they apply any customizations in SearchBuilder, by using RangeLimitSearchBuilder they have an additional place they need to copy them to?

Since RangeLimitSearchBuilder extends SearchBuilder, I wouldn't think you would need to make any changes.

jrochkind:
>  I'm not understanding. No more code or configuration is needed in this branch than is necessary on main.

In an application with more than one CatalogController-type controller, with different SearchBuilders. Which is a design I think BL means to support? 

Sorry, I tried to explain that above, but it's hard to talk about code. 

The actual CI test case only tests the equivalent of a freshly generated BL app with only one CatalogController and only one SearchBuilder, so no more code is needed in the test case, true.

jcoyne:
Yes, that makes sense, if you added a second controller with a separate SearchBuilder, then you'd also want to create a separate RangeLimitSearchBuilder that extends that other SearchBuilder.

jrochkind:
Previously, there was no extra work you had to do to install `blacklight_range_limit` no matter how many controllers you had. 

To me this is an undesirable change, fairly confusing to explain/document (or write fiddly code-generation tools for), resulting in a lot of extra boilerplate code that has to all match. 

I remember how much work some folks put into blacklight to make it possible to have more than one controller (searching different things), it was a priority feature at one point to make possible/easy.

PR #326, opened 2025-11-25
Explicitly support Blacklight 9 releases (not just pre-releases)

sandbergja:


sandbergja:
Yes, true!  The only thing that is absolutely necessary in this PR is updating the gemspec to allow 9.0.0 and above, not just prereleases (as the current `< 9` does in main).

PR #327, opened 2025-12-03
bump to version 9.2.0

jrochkind:


jrochkind:
My understanding is it's considered acceptable in community norms to review and merge your own PR when it is bumping a version number only. Let me know if there's disagreement, and I'll stop doing it in the future!

PR #335, opened 2026-07-01
Add inputmode=numeric to range fields

jcoyne:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inputmode#numeric

This helps UX on mobile safari, which otherwise shows the keyboard input.

### DIFF
--- a/app/components/blacklight_range_limit/range_input_component.html.erb
+++ b/app/components/blacklight_range_limit/range_input_component.html.erb
@@ -4,6 +4,7 @@
     @input_value,
     min: @min_value,
     max: @max_value,
+    inputmode: 'numeric',
     class: "form-control form-control-sm #{@css_class}" %>
 </div>
 


jrochkind:
We've definitely gone back and forth on this, we had it at one point and removed it. 

I forget if someone else asked for it to be removed or if it was my idea to remove before, let's see if I can find commit/PR from when it got removed last... I can't easily, predates current organization of files. 

I am willing to try it again... what does it do on desktop?  I think maybe the issue before was it took up too much horizontal space, but things have been moved about (no longer trying to fit submit button on same line) so I think that shouldn't be an issue.

jcoyne:
> I am willing to try it again... what does it do on desktop?

As far as I can tell, it has no effect on desktop browsers.

jrochkind:
> As far as I can tell, it has no effect on desktop browsers.

Huh weird, maybe this is a DIFFERENT attribute than we used to use, which I think added a 'spinner' control to the inputs, which some people found annoying/confusing/unwanted on desktop. 

That's definitely not happening for you?

Maybe that was a different attribute with value numeric, not inputmode!

